PR #60, opened 2019-03-14
chore: normalize ecs apis

menduz:


packages/build-ecs/index.ts
```diff
@@ -154,6 +154,10 @@ function emitFile(fileName: string, services: ts.LanguageService) {
 
     if (WATCH) {
       console.log('\nThe compiler is watching file changes...\n')
+    } else {
+      if (compiled.error) {
```
fmiras, line 158: Could we use `else if (compiled.error) {` here?

menduz: yes

packages/decentraland-ecs/src/decentraland/Components.ts
```diff
@@ -713,6 +713,7 @@ export class OnUUIDEvent<T extends keyof IEvents> extends ObservableComponent {
     }
 
     this.callback = callback
+    uuidEventSystem.handlerMap[this.uuid] = this
```
menduz: @pravusjif please move the line

```ts
uuidEventSystem.handlerMap[this.uuid] = this
```
back into the constructor of `OnChanged` and `OnTextSubmit`.

a comment can be added like
```ts
// this injection is necessary ONLY in events that are ALWAYS turned on and are
// not assignable to entities. Like events for the UI elements
uuidEventSystem.handlerMap[this.uuid] = this
```

packages/decentraland-ecs/src/decentraland/UIShapes.ts
```diff
@@ -10,7 +10,7 @@ export abstract class UIShape extends ObservableComponent {
    * Defines if the entity and its children should be rendered
    */
   @ObservableComponent.field
-  id: string | null = null
+  name: string | null = null
```
menduz, line 13: Why do we need a 'name' field? is this used somewhere?

pravusjif: We use the 'name' field (previously called 'id' and didn't provide any real use for the dev on typescript) mostly for debugging, so that by configuring a name to the components/entities, once instantiated in unity we can finde them quickly

packages/ui/avatar/avatarSystem.ts
```diff
@@ -9,8 +9,8 @@ import {
   Vector3,
   Quaternion,
   Component,
-  Scalar,
-  OnPointerDown
+  Scalar//,
+  //OnPointerDown
```
fmiras: Why is this a comment?

packages/ui/avatar/avatarSystem.ts
```diff
@@ -26,7 +26,7 @@ import {
 } from 'shared/comms/types'
 import { execute } from './rpc'
 import { ComponentGroup } from 'decentraland-ecs/src/ecs/ComponentGroup'
-import { showAvatarWindow, currentAvatarId, hideAvatarWindow } from './avatarWindow'
+//import { showAvatarWindow, currentAvatarId, hideAvatarWindow } from './avatarWindow'
```
fmiras: Another one here

packages/ui/avatar/avatarSystem.ts
```diff
@@ -146,36 +150,31 @@ export class AvatarEntity extends Entity {
   constructor(public name: string) {
     super(name)
 
-    const clicked = new OnPointerDown(this.clicked)
-
     {
       this.labelEntity.setParent(this)
-      this.labelEntity.getComponentOrCreate(Transform).position.y = 2
+
+      let labelTransform = this.labelEntity.getComponentOrCreate(Transform)
```
fmiras: Could this be `const` instead?

public/ecs-parcels/-200,110-relative-layout-ui/game.ts
```diff
@@ -1,58 +1,61 @@
...
+// TODO: UI-Migration
+// a bg.cornerRadius = 10
+bg.color = Color4.Black()
 
 // --- INVENTORY
 
-const inventoryContainer = new UIContainerStackShape(container)
+const inventoryContainer = new UIContainerStack(container)
 inventoryContainer.adaptWidth = true
 // inventoryContainer.adaptHeight = true --- you can only set adaptWidth (X)OR adaptHeight for it to work correctly
```
fmiras: Could we either delete this or put a TODO task with an issue attached?

public/ecs-parcels/-200,110-relative-layout-ui/game.ts
```diff
@@ -1,58 +1,61 @@
...
 inventoryContainer.width = '40%'
-inventoryContainer.top = '100px'
-inventoryContainer.left = '10px'
-inventoryContainer.color = 'white'
-inventoryContainer.background = 'blue'
+inventoryContainer.positionY = '-100px'
+inventoryContainer.positionX = '10px'
+// inventoryContainer.color = 'white' // previously set child text color
```
fmiras: Same here

packages/dcl/widgets/ui.ts
```diff
@@ -0,0 +1,58 @@
...
+        main,
+        data: {},
+        id,
+        mappings: []
+      },
+      context
+    )
+
+    this.context.rootEntity.name = this.context.rootEntity.id = id
+    // tslint:disable-next-line:no-unused-expression
```
fmiras, line 30: I think we should either respect this rule or remove it, there is no point of putting it and avoid it with comments

packages/shared/world/parcelSceneManager.ts
```diff
@@ -30,6 +30,19 @@ export function getParcelById(id: string) {
   return null
 }
 
+/**
+ * This function receives the list of { type: string, data: ILand } from a remote worker.
+ * It loads and unloads the ParcelScenes from the world
+ */
+export function getParcelById(id: string) {
+  for (let parcelSceneWorker of loadedParcelSceneWorkers) {
```
fmiras, line 38: We're using in some cases traditional `for` indexed-syntax and in other cases this syntax, I'd to be consistent with this, here we should use either es5 `for` statement or `Array.prototype.find` function

fmiras (dismissed):

packages/decentraland-ecs/src/decentraland/Components.ts
```diff
@@ -3,6 +3,7 @@ import { Vector3, Quaternion, Matrix, MathTmp, Color3 } from './math'
 import { AnimationState } from './AnimationState'
 import { newId } from '../ecs/helpers'
 import { IEvents } from './Types'
+import { uuidEventSystem } from './Systems';
```
matiasbargas: This import is causing a circular dependency because `System` is adding `Components` as import as well. I Recommend to move OnUUIDEvent to a new file to solve this.

matiasbargas (changes requested):

public/ecs-parcels/-200,110-relative-layout-ui/game.ts
BrianAmadori: Those comments are outdated, I removed all of them. We must create a task to fix this scene as its written for the old SDK though.

public/ecs-parcels/-200,110-relative-layout-ui/game.ts
BrianAmadori: Those comments are outdated, I removed all of them. We must create a task to fix this scene as its written for the old SDK though.

packages/shared/world/parcelSceneManager.ts
BrianAmadori: The traditional `for` indexed is o.k. for arrays, here we are dealing with a `Set`. Iteration-wise they are pears and oranges. The `of` approach is the most readable one with this data structure AFAIK.

packages/dcl/widgets/ui.ts
BrianAmadori: I believe this is a intended special case because the `new SceneWorker()` triggers its initialization and registers it to `ParcelScene`. You don't want to save it to a reference, just get it working.

packages/shared/world/parcelSceneManager.ts
menduz: It is ok. It doesn't have a .find function

packages/dcl/widgets/ui.ts
menduz: It is normally an error to have expressions in that way. Except for this case, where it is necessary. I believe it is ok to leave the comment and the rule

menduz:
@BrianAmadori hay que hacer un `export * from ...` en `decentraland-ecs/src/index.ts`

porque 

```ts
import { OnClick } from 'decentraland-ecs/src'
```
tiene que funcionar OOTB

matiasbargas (approved):

PR #61, opened 2019-03-15
feat: textures as disposable components

### LINKED ISSUE #22: Add texture as a disposable component
Because right now we only address textures by 'src'. We will want this in the future, the sooner we have it the better, that means less changes in the future.

menduz:
closes #22 
**DO NOT MERGE UNTIL IMPLEMENTED IN UNITY**
BLOCKED BY: https://github.com/decentraland/unity-client/issues/157


packages/engine/components/disposableComponents/Texture.ts
```diff
@@ -0,0 +1,87 @@
...
+  texture: BABYLON.Texture
+  meshObserver: Observer<{ type: string; object: BABYLON.TransformNode }>
+
+  constructor(ctx, uuid) {
+    super(ctx, uuid)
+    this.loadingDone = false
+  }
+
+  static async getFromComponent(context: SharedSceneContext, id: string): Promise<BABYLON.Texture> {
+    const component = context.getComponent(id)
```
BrianAmadori, line 23: If the texture is still being loaded as we call this, this wouldn't return a empty texture? From Unity side, this method's cousin waits for the texture load to finish before exiting.

menduz: Babylon first creates the texture object, then it loads the texture. Once the image data is loaded it is written to the texture buffer of the texture object. The reference to the object is the same always. Is there a way to do that in unity?

BrianAmadori: Using the common `Texture/Texture2D` class i don't think so, but it doesn't matter. It's already handled.

eordano:
Unlocked

PR #75, opened 2019-03-20
chore: json split for LoadParcelScenes message

BrianAmadori:
# What? 
This splits the LoadParcelScenes json with `<break>` statements. 

# Why? 
This allows throttling the JSON parsing from game engine side, as its too big now and causes performance hiccups.

### DIFF
--- a/packages/unity-interface/dcl.ts
+++ b/packages/unity-interface/dcl.ts
@@ -60,8 +60,22 @@ const unityInterface = {
   LoadParcelScenes(parcelsToLoad: LoadableParcelScene[]) {
     const parcelScenes = JSON.stringify({ parcelsToLoad })
     if (parcelScenes !== lastParcelScenesSent) {
+
       lastParcelScenesSent = parcelScenes
-      gameInstance.SendMessage('SceneController', 'LoadParcelScenes', parcelScenes)
+      var finalJson: string = "";
+
+      //NOTE(Brian): split json to be able to throttle the json parsing process in engine's side
+      for (var i = 0; i < parcelsToLoad.length; i++) {
+        var parcel = parcelsToLoad[i]
+        const json = JSON.stringify(parcel)
+
+        finalJson += json;
+
+        if (i < parcelsToLoad.length - 1)
+          finalJson += "<break>"
+      }
+
+      gameInstance.SendMessage('SceneController', 'LoadParcelScenes', finalJson)
     }
   },
   sendSceneMessage(parcelSceneId: string, method: string, payload: string) {
--- a/packages/unity-interface/dcl.ts
+++ b/packages/unity-interface/dcl.ts
@@ -66,17 +66,18 @@ const unityInterface = {
     if (parcelScenes !== lastParcelScenesSent) {
 
       lastParcelScenesSent = parcelScenes
-      var finalJson: string = "";
+      let finalJson: string = "";
 
       //NOTE(Brian): split json to be able to throttle the json parsing process in engine's side
-      for (var i = 0; i < parcelsToLoad.length; i++) {
-        var parcel = parcelsToLoad[i]
+      for (let i = 0; i < parcelsToLoad.length; i++) {
+        const parcel = parcelsToLoad[i]
         const json = JSON.stringify(parcel)
 
         finalJson += json;
 
-        if (i < parcelsToLoad.length - 1)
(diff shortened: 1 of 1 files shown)


packages/unity-interface/dcl.ts
```diff
@@ -64,8 +64,22 @@ const unityInterface = {
   LoadParcelScenes(parcelsToLoad: LoadableParcelScene[]) {
     const parcelScenes = JSON.stringify({ parcelsToLoad })
     if (parcelScenes !== lastParcelScenesSent) {
+
       lastParcelScenesSent = parcelScenes
-      gameInstance.SendMessage('SceneController', 'LoadParcelScenes', parcelScenes)
+      var finalJson: string = "";
```
fmiras: It is not recommended to use `var` statements anymore since ES6, use `let` variables that are going to be redefined and `const` for variables that are not going to be redefined

packages/unity-interface/dcl.ts
```diff
@@ -64,8 +64,22 @@ const unityInterface = {
...
+
+      //NOTE(Brian): split json to be able to throttle the json parsing process in engine's side
+      for (var i = 0; i < parcelsToLoad.length; i++) {
+        var parcel = parcelsToLoad[i]
```
fmiras: You can also use `const` here

packages/unity-interface/dcl.ts
```diff
@@ -64,8 +64,22 @@ const unityInterface = {
...
+        const json = JSON.stringify(parcel)
+
+        finalJson += json;
+
+        if (i < parcelsToLoad.length - 1)
```
fmiras: Always use curly braces " { } " for defining blocks. It helps code readability + avoid skipping code execution

fmiras (changes requested):
We're still sending just one message, have you got one example of the JSON that you want to parse? maybe you can split it without a "<break>" delimiter (wtf, i can edit your comment?)

BrianAmadori:
> We're still sending just one message, have you got one example of the JSON that you want to parse? maybe you can split it without a "" delimiter (wtf, i can edit your comment?)

It should look like

```
{
   { stuff },
   { stuff }
}<break>
{
   { stuff },
   { stuff }
}<break>
{
   { stuff },
   { stuff }
}
```

I have to admit that I kind of hate the approach. I couldn't split it in several messages because the client expect the message to be processed instantly and this doesn't work well in WSS mode. I used the `<break>` as delimiter because I don't feel comfy with a single character. Maybe its better to just use `chr(13)` or similar.

fmiras:
> > We're still sending just one message, have you got one example of the JSON that you want to parse? maybe you can split it without a "" delimiter (wtf, i can edit your comment?)
> 
> It should look like
> 
> ```
> {
>    { stuff },
>    { stuff }
> }<break>
> {
>    { stuff },
>    { stuff }
> }<break>
> {
>    { stuff },
>    { stuff }
> }
> ```
> I have to admit that I kind of hate the approach. I couldn't split it in several messages because the client expect the message to be processed instantly and this doesn't work well in WSS mode. I used the `<break>` as delimiter because I don't feel comfy with a single character. Maybe its better to just use `chr(13)` or similar.

What if we use the expression `}\n{` or `}{`? That's not valid JSON so we can detect the different messages like that and save 7 extra bytes string per message (<break>)

BrianAmadori:
> > > We're still sending just one message, have you got one example of the JSON that you want to parse? maybe you can split it without a "" delimiter (wtf, i can edit your comment?)
> > 
> > 
> > It should look like
> > ```
> > {
> >    { stuff },
> >    { stuff }
> > }<break>
> > {
> >    { stuff },
> >    { stuff }
> > }<break>
> > {
> >    { stuff },
> >    { stuff }
> > }
> > ```
> > I have to admit that I kind of hate the approach. I couldn't split it in several messages because the client expect the message to be processed instantly and this doesn't work well in WSS mode. I used the `<break>` as delimiter because I don't feel comfy with a single character. Maybe its better to just use `chr(13)` or similar.
> 
> What if we use the expression `}\n{` or `}{`? That's not valid JSON so we can detect the different messages like that and save 7 extra bytes string per message ()

Not bad, only thing is JSON is gonna look like a bit funky:
```
{
   { stuff },
   { stuff }
}}{
{
   { stuff },
   { stuff }
}
```
But the split would work.

fmiras (approved):

Issue #28, opened 2019-01-29
Animation looping can't be stopped in 5.1


nearnshaw:
If I set the `looping` property of an AnimationClip to false, it changes nothing, it keeps looping the animation.

```ts
// Create Gnark
let gnark = new Entity()
gnark.add(new Transform({
 position: new Vector3(5, 0, 5),
 scale: new Vector3(0.75, 0.75, 0.75)
}))
gnark.add(new GLTFShape('models/gnark.gltf'))
engine.addEntity(gnark)

let animator = new Animator()

// // Add walk animation
const walkClip = new AnimationClip("walk")
walkClip.looping = false
// I tried this too, same problem:
//walkClip.setParams({looping:false})

animator.addClip(walkClip)
gnark.add(animator)

walkClip.play()
```

menduz:
Verified, it happens as written. Unfortunately it is super hard to fix without hacking babylon.
Workaround: 
```ts
clip.looping = false 
clip.stop()
clip.play()
```

PR #84, opened 2019-03-21
feat: reset animation

### LINKED ISSUE #51: Add a `reset()` to animation clips
Today to reset an animation I have to delete the previous one and create another:
```ts
// It lasts 2 seconds
const walkClip = new AnimationClip('walk')
walkClip.play()
setTimeout(() => walk.pause(), 1000)
// reset
const walkClip2 = new AnimationClip('walk')
// Use the new animation
walkClip2.play()
```

Developers should be able to reset existing ones:
```ts
const walkClip = new AnimationClip('walk')
walkClip.play()
setTimeout(() => walkClip.pause() && walkClip.reset(), 1000)
// Use it again
walkClip.play()
```

fmiras:
Closes #51 + fix animation pausing which didn't maintain the clip state after resuming it again

### DIFF
--- a/packages/decentraland-ecs/src/decentraland/AnimationClip.ts
+++ b/packages/decentraland-ecs/src/decentraland/AnimationClip.ts
@@ -50,6 +50,13 @@ export class AnimationClip extends ObservableComponent {
   @ObservableComponent.field
   public speed: number = defaultParams.speed
 
+  /**
+   * Has the animation been reset after starting? default: false
+   */
+  // @internal
+  @ObservableComponent.field
+  public resetSinceStart: boolean = false
+
   // @internal
   @ObservableComponent.readonly
   readonly name: string = newId('AnimClip')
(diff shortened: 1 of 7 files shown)


nearnshaw:
I tested this, there are still some issues to work out here.  I made the scene on (-100, 111) more complete to test it better. Here are the use cases that failed:

Reproduce the following steps:
Play the animation, push pause at a random
then push play again
expected behavior:  The animation should play from where it was left off
actual behavior:  The animation plays from the start

Reproduce the following steps:
Play the animation, push stop at random
expected behavior:  The animation should return to frame 1 and stop playing
actual behavior:  The animation stops playing at the place where it was. It works exactly the same as pause.

Reproduce the following steps:
Play the animation, push pause at random. Then push stop
expected behavior:  The animation should return to frame 1 and continue paused
actual behavior:  The animation does nothing.

Also, setting animation looping can't be stopped, but that's a known bug tracked by https://github.com/decentraland/explorer/issues/28

fmiras:
Too much state change, we're closing this and redo it again.

PR #106, opened 2019-04-08
docs: add EventManager examples

### LINKED ISSUE #12: chore: Implement user scene post-initialization event
It should be an event fired when the game.ts code has run and the scene is ready to render stuff.

menduz:
Add a EventManager example.
Adds scene created event, closes #12

### DIFF
--- a/packages/decentraland-ecs/src/decentraland/Implementation.ts
+++ b/packages/decentraland-ecs/src/decentraland/Implementation.ts
@@ -42,10 +42,8 @@ export class DecentralandSynchronizationSystem implements ISystem {
     this.dcl.onEvent(event => {
       switch (event.type) {
         case 'uuidEvent':
-          const e = new UUIDEvent()
-          e.uuid = (event.data as any).uuid
-          e.payload = (event.data as any).payload
-          engine.eventManager.fireEvent(e)
+          const data = event.data as any
+          engine.eventManager.fireEvent(new UUIDEvent(data.uuid, data.payload))
           break
       }
     })
--- a/packages/decentraland-ecs/src/ecs/EventManager.ts
+++ b/packages/decentraland-ecs/src/ecs/EventManager.ts
@@ -1,4 +1,4 @@
-import { error } from './helpers'
+import { error, newId } from './helpers'
 
 const eventNameSymbol = '__event_name__'
 
@@ -20,11 +20,16 @@ function getEventNameFromConstructor<T>(ctor: IEventConstructor<T>): string {
   return (ctor[eventNameSymbol] as any) as string
 }
 
+type EventListener<X> = {
+  listener: X
+  fn: (this: X, event: any) => void
+}
+
 /**
  * @public
  */
 export class EventManager {
-  private listeners: Record<string, Array<{ listener: any; fn: (event: any) => void }>> = {}
+  private listeners: Record<string, Array<EventListener<any>>> = {}
 
   addListener<T, X>(eventClass: IEventConstructor<T>, listener: X, listenerFunction: (this: X, event: T) => void) {
     if (!eventClass || typeof (eventClass as any) !== 'function') {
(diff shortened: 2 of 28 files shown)


packages/decentraland-ecs/src/decentraland/Implementation.ts
```diff
@@ -42,10 +42,8 @@ export class DecentralandSynchronizationSystem implements ISystem {
     this.dcl.onEvent(event => {
       switch (event.type) {
         case 'uuidEvent':
-          const e = new UUIDEvent()
-          e.uuid = (event.data as any).uuid
-          e.payload = (event.data as any).payload
-          engine.eventManager.fireEvent(e)
+          const data = event.data as any
```
fmiras, line 45: Wouldn't be cooler to make
```ts
const { data }: any = event
```
so we cast `event.data` object to `any` type once?

test/unit/ecs.test.tsx
```diff
@@ -1071,6 +1062,18 @@ describe('ECS', () => {
...
+      const materials: PBRMaterial[] = []
+
+      for (let [, x] of context.disposableComponents) {
+        if (x instanceof PBRMaterial) {
+          materials.push(x)
+        }
+      }
+
+      return materials
+    }
```
fmiras, line 1075: Just a comment, this function could be lot more readable if separate them in lambda functions like this:

```ts
function getMaterials(context: SharedSceneContext): DisposableComponent[] {
	return Array.from(context.disposableComponents)
    	.map(pair => pair[1]) // or .map(([, component]) => component)
        .filter(component => component instanceof PBRMaterial)
}
```

or even just asking for the entries:
```ts
function getMaterials(context: SharedSceneContext): DisposableComponent[] {
	return Array.from(context.disposableComponents.entries())
        .filter(component => component instanceof PBRMaterial)
}

```

packages/decentraland-ecs/src/ecs/EventManager.ts
```diff
@@ -96,19 +101,18 @@ export class EventManager {
...
   if (takenEventNames.indexOf(eventName) !== -1) {
     throw new Error(`The event name ${eventName} is already taken`)
   }
 
   takenEventNames.push(eventName)
 
-  return (target: any) => {
-    target[eventNameSymbol] = eventName
+  return <TFunction extends Function>(target: TFunction): TFunction | void => {
+    // tslint:disable-next-line:semicolon
```
fmiras, line 114: Not related to the PR but what do you think of removing this rule? prettier already remove unnecessary semi-colons for us and we already have 16 matching results of this comment. Is not bad to use it

fmiras (changes requested):
LGTM but I'd to discuss the comments I left

packages/decentraland-ecs/src/decentraland/Implementation.ts
menduz: That does not produce the same result

packages/decentraland-ecs/src/ecs/EventManager.ts
menduz: Sure, we should wait to merge the open things before, the merge conflicts will kill us

test/unit/ecs.test.tsx
menduz: Is it that bad? I like functional thing when they save us dozens of lines, without sacrificing readability. This one is pretty clear

nearnshaw:
@menduz  

I'm documenting events right now.
Will we keep the second argument of the `addListener` function?  I think I remember a conversation we had where you said that was pointless and you'd try to remove it.

```
events.addListener(UpdateEvent, null, ({ entity, dt }) => {
    // function
})
```

Will it remain there for the release?

menduz:
Yes, it will remain for performance reasons.

PR #112, opened 2019-04-12
feat: use content.decentraland.today on .today domains

fmiras:


menduz:
It is inconsistent with the rest of the configurations, because it will be fetching content from the domain, but in .zone you can choose to use mainnet, but the content will be fetched from .zone which is not sound

fmiras:
> It is inconsistent with the rest of the configurations, because it will be fetching content from the domain, but in .zone you can choose to use mainnet, but the content will be fetched from .zone which is not sound

We've already discussed this at slack a few times, before creating the PR I asked if we were going to prior the web3 provider or the domain and for consistency reasons (e.g. decentraland marketplace priors domain, so even if you select mainnet at .zone it will display ropsten data) we decided to go forward with the domains approach

menduz (approved):

m-sossich (approved):

PR #115, opened 2019-04-17
feat: add reset and stop functions to animation clips

fmiras:
Closes #84


menduz:
When does you change `shouldReset` to false to avoid continuous resets?

fmiras:
> When does you change `shouldReset` to false to avoid continuous resets?

Good catch, I didn't test it yet. My bad.

menduz:
when do you reset that flag from the SDK?

menduz (approved):

PR #118, opened 2019-04-22
feat: create new configuration approach

fmiras:
# What? 
This PR cleans some unused configuration, make the explorer fetch DCL contracts from [contracts.decentraland.org](https://contracts.decentraland.org) and starts using a full TLD-based approach for all the server/backend decentraland services

Also it fixes the communications server connection (was not working at some environments)

**NOTE:** Give feedback of this global config approach (there are some new global variables that must be set now at `initShared` function with the new `setNetwork` function)

### DIFF
--- a/packages/config/index.ts
+++ b/packages/config/index.ts
@@ -10,17 +10,11 @@ export const performanceConfigurations = [
 
 export const NETWORK_HZ = 10
 
-export const playerSphereRadius = 0.4
-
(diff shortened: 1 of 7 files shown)


fmiras:
> This is harder to read than
> 
> ```ts
> const inviteAddress = decentralandConfigurations.invite
> ```

Changed

Issue #25, opened 2019-01-30
An event to say when an animation finished playing


nearnshaw:
It would be very useful to know when an AnimationClip has gone through the whole animation and finished. It's useful both for looping and non-looping animations.

For a use case see this tutorial:
https://decentraland.org/blog/tutorials/motion-animations-in-SDK-5-part-2

Here I wanted to pause the character while he played a TurnRight animation on each corner of his path. The work around I found was to create a timer to freeze everything for just as long as the animation lasts. This workaround isn't perfect, it's sometimes jumpy. 

It would be ideal to have some kind of event listener, like we have for clicks. The listener should have the AnimationClip instance as an input when instancing it, and so the listener only works for one instance of an AnimationClip

let turnRight_animation = new AnimationClip("turnRight")

animation.subscribe(turnRight_animation, e => {
  log("turned right", e)
})

PR #122, opened 2019-04-24
feat: create onAnimationEnd event

### LINKED ISSUE #28: Animation looping can't be stopped in 5.1

fmiras:
Closes #28 and #25

### DIFF
--- a/packages/decentraland-ecs/src/decentraland/AnimationClip.ts
+++ b/packages/decentraland-ecs/src/decentraland/AnimationClip.ts
@@ -74,6 +74,14 @@ export class AnimationClip extends ObservableComponent {
     this.speed = params.speed || this.speed
   }
 
+  toJSON() {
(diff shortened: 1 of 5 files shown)


menduz (changes requested):
1. At least one test is required.
2. We need to hook the event to the animation somehow (in the SDK)

PR #128, opened 2019-04-25
feat: kill BUTTON_A_* in favor of BUTTON_*

eordano:


fmiras:
There are test parcels that are still using `BUTTON_A_*` we still have to update those.

menduz:
This is because pointers have several buttons.

A single pointer (mouse) have button A and B (left and right click)

A VR controller has several (A,B, Trigger)

The main (and only) pointer of the mobile has two buttons A and B

PR #130, opened 2019-04-25
feat: add unityPreview.ts entry point

menduz:
It adds a separate preview for unity builds. I'd like to copy this file too to the `artifacts` so we can switch to unity in the CLI to compare progress and features. Leaving Babylon as main entry point @eordano @fmiras

# What? 
...

# Why? 
...


menduz:
Mergeamo?

fmiras (approved):

PR #131, opened 2019-04-25
[WIP] feat: fix enter focus on chat (babylon)

menduz:
Merge after releasing Babylon version

TODO:
- Make focus an event, decouple from babylon
- Create UIController API to handle this events?


fmiras:
Won't fix for babylon only for unity

PR #133, opened 2019-04-26
feat: kill BUTTON_A_* in favor of BUTTON_* (#128)

fmiras:
This reverts commit 456162d6e3242308b399c71797b88372d41e61f0.

We will merge this only when we have better product definition

Reference #128


menduz (approved):

menduz:
The definition so far is: 
- It has to be consistent with the documentation
- We may add information about the button in the event, de-normalizing events is not a good practive

PR #140, opened 2019-05-01
fix: children that dont have material dont count

eordano:
A GLTF had a camera entity that was causing it to show as "fail" with impossible-to-debug conditions.


packages/engine/entities/BaseEntity.ts
```diff
@@ -430,6 +430,7 @@ export class BaseEntity extends BABYLON.AbstractMesh {
     let min = boundingInfo.boundingBox.minimumWorld
     let max = boundingInfo.boundingBox.maximumWorld
     for (let i = 1; i < children.length; i++) {
+      if (!children[i].material) continue
```
menduz, line 434:
```suggestion
      // Cameras/lights doesn't have any materials and may be outside of the scene.
      if (!children[i].material) continue
```

menduz (approved):

Issue #98, opened 2019-04-03
Integrate analytics


matiasbargas:
Track and send metrics listed in the document to segment.  https://docs.google.com/document/d/1e_6s5wTw08V7BuBI-0RlRoDk8dLUgOV82IKZNmO_Yrw/edit#heading=h.tneub8dv35mt

PR #146, opened 2019-05-02
feat: add segment support

fmiras:
Contributes #98 

Can't identify users until identity integration/definitions


eordano (approved):

packages/shared/comms/peers.ts
```diff
@@ -163,6 +164,7 @@ export function receiveUserVisible(uuid: string, visible: boolean) {
     uuid,
     visible
   })
+  queueTrackingEvent('User visibility', { uuid, visible })
```
menduz: Following the `Single responsibility principle` it would be better to have a file called `analytics` or something like that, listening to `avatarMessageObservable`.

It will decouple this file from analytics completely

menduz (changes requested):

packages/shared/world/positionThings.ts
```diff
@@ -31,18 +32,35 @@ positionObservable.add(event => {
...
   const gridPosition = Vector2.Zero()
+  let previousWorldPosition: ReadOnlyVector3 | null = null
 
   function updateUrlPosition(cameraVector: ReadOnlyVector3) {
+    if (previousWorldPosition === null) {
+      previousWorldPosition = { x: cameraVector.x, y: cameraVector.y, z: cameraVector.z }
+    }
+
+    if (seconds === 10 || distanceTraveled > 50) {
+      queueTrackingEvent('User Position', { position: cameraVector, distance: distanceTraveled })
```
menduz: Here too, we have a position observable. You can decouple the tracking code from the implementation by listening to the position observable and keep this implementation clean and decoupled from tracking

menduz (changes requested):

packages/ui/avatar/chatWindow.ts
```diff
@@ -13,9 +13,11 @@ import {
   UIScrollRect
 } from 'decentraland-ecs/src/decentraland/UIShapes'
 
+import { MessageEntry } from 'shared/types'
+import { queueTrackingEvent } from 'shared/analytics'
```
menduz: this doesn't work since packages/ui is a web worker

packages/shared/analytics.ts
```diff
@@ -0,0 +1,113 @@
...
+
+chatObservable.add((event: any) => {
+  if (event.type === ChatEvent.MESSAGE_RECEIVED) {
+    queueTrackingEvent('Chat message received', { lenght: event.data.message.lenght })
+  } else if (event.type === ChatEvent.MESSAGE_SENT) {
+    queueTrackingEvent('Send chat message', { messageId: event.data.messageId, length: event.data.message.length })
+  }
+})
+
+avatarMessageObservable.add(({ type, ...data }) => queueTrackingEvent(type, data))
```
menduz: I'd wrap this `observables.add` into functions like `function hookAnalytics() {` and then only call them inside the function `initialize` to avoid side effects at startup

menduz (changes requested):

packages/ui/avatar/chatWindow.ts
fmiras: I just removed it, created an issue #180 to make it in another PR

PR #147, opened 2019-05-03
feat: nft shape

belohlavek:
# What? 
NFT Shape with picture frame

# Why? 
One small step for man...


eordano (changes requested):
I don't see any validation logic here

menduz:
It is on purpose, because we don't have accounts yet for the builder, no NFT could be shown in the builder if we add it.

menduz (approved):

Issue #148, opened 2019-05-06
Improve DX by making compile process faster


fmiras:
# What? 

Try by using `ncc`, plain webpack or changing the decentraland-compiler approach. Refactor makefile to avoid rebuilding already built packages

# Why?

Waiting for compilation + tests takes **a lot** of time and slows the project development

menduz:
We compile almost 100 projects with different configurations.. at this scale the only viable option is to manually tune the compiler, which is plain webpack.

ncc is for a different use case, we should remove it from the equation

PR #167, opened 2019-05-11
chore: only build modified ECS projects

fmiras:
Contributes to #148 

# What? 
ECS Projects only are going to build in case of modification (even at first `make build`)

**TODO:** 
- [x] Calculate stat of child folder files

### DIFF
--- a/scripts/buildECSprojects.ts
+++ b/scripts/buildECSprojects.ts
@@ -1,14 +1,8 @@
-// tslint:disable:no-console
-// this file uses console.log because it is a helper
-
-import glob = require('glob')
-import path = require('path')
-import { dirname } from 'path'
+import * as path from 'path'
 import { spawn } from 'child_process'
 
-const folders = glob
-  .sync(path.resolve(__dirname, '../public/ecs-parcels/*/game.ts'), { absolute: true })
-  .map($ => dirname($))
+import * as fs from 'fs-extra'
+import glob = require('glob')
 
 const PWD = process.cwd()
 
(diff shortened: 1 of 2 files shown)


scripts/buildECSprojects.ts
```diff
@@ -19,10 +13,43 @@ const env = {
...
-    env,
-    stdio: 'inherit'
-  }).ref()
-})
+const folders = glob
+    .sync(path.resolve(__dirname, '../public/ecs-parcels/*/game.ts'), { absolute: true })
+    .map($ => path.dirname($))
+
+  // tslint:disable-next-line: no-floating-promises
+;(async () => {
```
menduz: it is better to do 
```ts
async function main() { ... }

main().catch(err => {
  console.error(err)
  process.exit(1)
})

eordano:
why is this blocked?

eordano (approved):

fmiras:
> why is this blocked?

It was blocked because we're only stating the parents folder, that means if we change something in `game.ts` it wasn't going to compile eiter. Now that's solved

menduz (changes requested):
It breaks the watch mode `node ./scripts/buildECSprojects.js --watch`

eordano:
Is this still blocked? What's the cause of it breaking the watcher?

fmiras:
> Is this still blocked? What's the cause of it breaking the watcher?

Because in case it doesn't have to rebuild it, it just skip `decentraland-compiler` which is the one in charge of watching those files in the first place

I'll close this PR. We should either implement this logic in `decentraland-compiler` as a `--cache` flag or implement TS 3.4 incremental builds

PR #174, opened 2019-05-13
chore: reenable new scene lifecycle

fmiras:
This reverts commit ea6565c5934b4a3dcc2aec1ca466a91713d913ab.


eordano (approved):

pravusjif:
The client is sending the wrong "parcel id".
Before, it sent the base coordinates as the parcel id (e.g. the Floating World scene's base coords are "20,20"), but now it is sending the last scene's parcel coords as the scene id (now the Floating World sends "23,27").

From what I could debug in the **parcelSceneManager.ts** in **getParcelById()**, the id is already wrong at that point in time when that function is called.

Besides that, it's working fine

eordano:
Updated!

eordano:
@pravusjif 

I modified the name of that function but I think I need some help on how are you using that "id".

I am now exposing these two functions:
getSceneWorkerByCID
getSceneWorkerByBaseCoordinates

I'm not sure which is the one you need from the unity interface. Please verify it!

eordano:
(this is the key change: https://github.com/decentraland/explorer/pull/174/files#diff-0370cbd3df4c427a40a107cc4ea4195a)

PR #184, opened 2019-05-16
fix: add timestamp field to tracking events

fmiras:
# What? 
add timestamp field to tracking events

# Why? 
because they are being queued to be sent to segment serially and the original segment timestamp could differ from the real action execution timestamp


menduz (approved):

menduz:
ping @dnul

Issue #202, opened 2019-05-28
Using setParent causes the triangles count to double


cazala:
**Version**: This affects `decentraland@3.0.2` and `decentraland-ecs@6.0.4`

Once an entity is added as a child of another entity, the triangle count for that entity duplicates. Here I will explain how to reproduce this bug, and I added a zip file at the end containing example project to reproduce it:

1. I created a beautiful scene on the builder, and the triangle count is `767`:

<img width="1068" alt="beautiful scene" src="https://user-images.githubusercontent.com/2781777/58507218-9b61c000-8167-11e9-80f4-087ddb18e6f7.png">

2. I exported that scene and opened it with the cli using `dcl start`, the triangle count remains `767`:

```ts
const floorBaseGrass_01 = new Entity()
const gltfShape = new GLTFShape(
  'models/FloorBaseGrass_01/FloorBaseGrass_01.glb'
)
floorBaseGrass_01.addComponentOrReplace(gltfShape)
const transform = new Transform({
  position: new Vector3(8, 0, 8),
  rotation: new Quaternion(0, 0, 0, 1),
  scale: new Vector3(1, 1, 1)
})
floorBaseGrass_01.addComponentOrReplace(transform)

const treeRoundPine_02 = new Entity()
const gltfShape_2 = new GLTFShape(
  'models/TreeRoundPine_02/TreeRoundPine_02.glb'
)
treeRoundPine_02.addComponentOrReplace(gltfShape_2)
const transform_2 = new Transform({
  position: new Vector3(8.5, 0, 7.5),
  rotation: new Quaternion(0, 0, 0, 1),
  scale: new Vector3(1, 1, 1)
})
treeRoundPine_02.addComponentOrReplace(transform_2)

// add entities
engine.addEntity(floorBaseGrass_01)
engine.addEntity(treeRoundPine_02)
```

<img width="1440" alt="still beautiful scene" src="https://user-images.githubusercontent.com/2781777/58507172-8127e200-8167-11e9-93d7-cc58ca35574d.png">

3. I wanted to rotate my scene clockwise, so I added a parent entity, rotated it, and set it as parent of the tree and ground entities:

```ts
const parent = new Entity()
const clockwise = new Transform({
  position: new Vector3(0, 0, 16),
  rotation: Quaternion.Euler(0, 90, 0)
})
parent.addComponent(clockwise)
engine.addEntity(parent)

floorBaseGrass_01.setParent(parent)
[...]

PR #203, opened 2019-05-28
fix: #202

menduz:


nearnshaw:
@menduz 
This also fixes #311 from Unity Client, right?  It looks like a duplicate issue
https://github.com/decentraland/unity-client/issues/311

Or is this a Babylon-only fix?    (or maybe this was never really an issue on Unity?)

menduz:
yes, it is the same

PR #224, opened 2019-06-05
feat: use the new & faster content service endpoint

eordano:
Before this change, the data download for a random location in the world is about ~24mb
image

Afterwards, it's 13.6mb
image

This is related to the latest content-service change.


packages/decentraland-loader/lifecycle/controllers/scene.ts
```diff
@@ -76,23 +77,30 @@ export class SceneLifeCycleController extends EventEmitter {
...
-          if (!this.futureOfPositionToCID[pos]) {
-            return
-          }
-          if (this.futureOfPositionToCID[pos].isPending) {
-            this.futureOfPositionToCID[pos].resolve(sceneCID)
-          }
-        }
-      })
+      try {
+        this.downloadManager
```
menduz: use await instead of .then .catch here

eordano: That's much better, thanks

packages/decentraland-loader/lifecycle/controllers/scene.ts
```diff
@@ -76,23 +77,27 @@ export class SceneLifeCycleController extends EventEmitter {
     }
     if (!this.futureOfPositionToCID[position]) {
       this.futureOfPositionToCID[position] = future<string>()
-      this.downloadManager.getParcelData(position).then(land => {
+      try {
+        const land = await this.downloadManager.getParcelData(position)
         if (!land) {
-          return
+          this.futureOfPositionToCID[position].reject(new Error(`Scene for position ${position} not found`))
```
menduz:
```suggestion
          throw new Error(`Scene for position ${position} not found`)
```

packages/decentraland-loader/lifecycle/controllers/scene.ts
```diff
@@ -76,23 +77,27 @@ export class SceneLifeCycleController extends EventEmitter {
...
+          return ''
```
menduz:
```suggestion

```

eordano: It was really hard to make this decision. It seems to me that failing to fetch a scene is better than a sentinel value, but not very far... (exceptions are for exceptional cases...)

I was thinking that changing the return type to `string|null` or even `string?` might be a better approach. Thoughts?

menduz (changes requested):
The only thing I see missing is to generate a new "scenes" file in the folder `public/local-ipfs` as long as "mappings"

eordano:
Yep, that's a blocker

packages/decentraland-loader/lifecycle/controllers/scene.ts
menduz: If you are thinking about a `Maybe<string>` monad, what should we do with the `None` results. Why do we add them to the equation?, because today rejections are clear, it breaks every flow with `await requestSceneCID(...)`. Is there anything useful in returning a null to that await instead of throw?

packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -1,73 +1,122 @@
-import { ILand, MappingsResponse, IScene, normalizeContentMappings } from 'shared/types'
+import { ILand, IScene, ParcelInfoResponse } from 'shared/types'
 import { jsonFetch } from 'atomicHelpers/jsonFetch'
-import { error } from 'util'
+import { error, log } from 'util'
+import { future, IFuture } from 'fp-future'
```
fmiras: As part of decentraland standars, we import first third party packages and then local packages (this should be at the top)

packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -1,73 +1,122 @@
...
-      this.options.contentServer + `/mappings?nw=${nw[0]},${nw[1]}&se=${nw[0] + 1},${nw[1] - 1}`
+      this.options.contentServer + `/scenes?x1=${nw[0]}&x2=${nw[0]}&y1=${nw[1]}&y2=${nw[1]}`
     )
 
     if (!responseContent.ok) {
       error(`Error in ${this.options.contentServer}/mappings response!`, responseContent)
-      throw new Error(`Error in ${this.options.contentServer}/mappings response!`)
+      const ret = new Error(`Error in ${this.options.contentServer}/mappings response!`)
+      promised.reject(ret)
+      throw ret
```
fmiras, line 39: Why are we both rejecting a promises and throwing and error? Shouldn't we doing one of both? is like executing a promise and it will break before even defining the `.catch()`

packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -1,73 +1,122 @@
...
     } else {
-      const contents = (await responseContent.json()) as MappingsResponse[]
+      const contents = (await responseContent.json()) as SceneMappingResponse[]
       if (!contents.length) {
+        promised.resolve(null)
         return null
```
fmiras, line 44: Here's the same, why do we resolve the promise if we're then returning null? I think is useless logic here

packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -1,73 +1,122 @@
...
+      log(`Resolved ${rootCID} already -- returning`)
+      return promised
+    }
+
+    const data: ILand = {
+      baseUrl,
+      scene,
+      mappingsResponse: content.content
+    }
+    log(`Mapping for ${sceneJsonMapping.hash} is ${JSON.stringify(data)}`)
```
fmiras: Can we `log.debug` this? The console is already too spammed by other process. I think we should start filtering

fmiras (dismissed):
I'd like to know why are we using lot's of more futures that promises, If it could be easily get done with promises, I suggest going on that direction is simplier, more standard-ish and let new developers understand the code easier

packages/decentraland-loader/lifecycle/controllers/scene.ts
eordano: I'd rather have a `Maybe<string>` than a string that is not a valid Scene CID. I would like to be conservative in what this function returns, and there *will* be parcels without content...

packages/decentraland-loader/lifecycle/controllers/download.ts
eordano: The `reject` is for all other callers to the function, the `throw` (or `return` in other places) is for the caller that triggered the fetch. This is to handle concurrent requests.

eordano:
> I'd like to know why are we using lot's of more futures that promises, If it could be easily get done with promises, I suggest going on that direction is simplier, more standard-ish and let new developers understand the code easier

I agree with this in principle. It seems to be a pattern that we're using a lot. Maybe it'd be better to just memorize the function -- that will simplify the code a lot. I'll take a swing at it later...

eordano:
Blocked by https://github.com/decentraland/cli/issues/440

menduz:
It also requires an implementation in this same repository. Look at the file: https://github.com/decentraland/explorer/blob/master/scripts/createMockJson.ts

menduz (dismissed):
It also requires an implementation in this same repository. Look at the file: https://github.com/decentraland/explorer/blob/master/scripts/createMockJson.ts

packages/decentraland-loader/lifecycle/controllers/scene.ts
eordano: No, race condition on multiple requests

eordano:
Ready to review
Blocked merge on:
- https://github.com/decentraland/content-service/pull/153
- https://github.com/decentraland/cli/pull/442

packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -1,7 +1,8 @@
+import { future, IFuture } from 'fp-future'
+
 import { ILand, IScene, ParcelInfoResponse } from 'shared/types'
 import { jsonFetch } from 'atomicHelpers/jsonFetch'
-import { error, log } from 'util'
-import { future, IFuture } from 'fp-future'
+import { error, debug } from 'util'
```
menduz: use our logger, not util

menduz (dismissed):
Does it write a `scenes` file?

eordano:
still needs more work

test/testHelpers.ts
```diff
@@ -392,57 +391,6 @@ export function loadTestParcel(
...
-
-        return mockedScene
-      }
-    } catch (e) {
-      error(`Error loading mock for ${parcel.parcel_id}`, e.message)
-    }
-  }
-  return null
-}
-
```
menduz, line 445: ❤️

eordano:
> Does it write a `scenes` file?

yes

> use our logger, not util

done

menduz:
EXCELLENT VVORK We are good to merge until we release the CLI with `/scenes`, because @next will not work with the current version of the CLI

menduz (approved):

fmiras (approved):

PR #226, opened 2019-06-05
[WIP] feat: different registries for different DARs

eordano:
Different DARs could point to different SchemaAPIs


eordano:
This was a PoC, Closing

Issue #243, opened 2019-06-11
Refactor entrypoints to include as less as possible debugging features on a production build


fmiras:
There are several development flags/features that are included in a production build such us `DEBUG`, `DEBUG_MESSAGES`, `DISABLE_AUTH`, `DEBUG_METRICS` URL flags and things like the WS connection for using Unity Editor, preview logic on explorer `tetra.html`

eordano:
@fmiras feel free to reopen if this is not fixed yet

PR #245, opened 2019-06-12
refactor: entrypoints

fmiras:
Contributes #243 

# What? 
- Correctly use `decentraland-auth` iframe (no more popup)
- Only create a WS connection on localhost debugging (moved to another entrypoint + delegated to `test.ts` server of `make watch`)
- Make `index.html` Unity no more `tetra.html`, moved babylon to `babylon.html`
- General refactor


packages/unity/DCL.ts
```diff
@@ -0,0 +1,33 @@
+import { initializeEngine } from './init'
+import { UnityGame } from './types'
+
+/**
+ * This is the global-scoped object that Unity executes to send `browserInterface`
+ * messages to Explorer
+ */
+export default class DCL {
```
menduz: No need to use a class, use namespaces or file-scoped variables

packages/unity/preview/DCLPreview.ts
```diff
@@ -0,0 +1,72 @@
...
+import { initializePreview } from './initializePreview'
+import DCL from '../DCL'
+
+declare var global: any
+
+/**
+ * This is the global-scoped object that Unity executes to send `browserInterface`
+ * messages to Explorer
+ */
+export default class DCLPreview extends DCL {
```
menduz: Why a class?

scripts/test.ts
```diff
@@ -169,6 +169,21 @@ function checkDiff(imageAPath: string, imageBPath: string, threshold: number, di
...
 
+  app.get('/', (req, res) => {
+    fs.readFile(resolve(__dirname, '../static/index.html'), 'utf8', (err, data) => {
+      if (err) {
+        res.send('There was an error reading index.html')
+        return
+      }
+
+      if (req.url.includes('ws') || req.url.includes('DEBUG')) {
+        res.send(data.replace('dist/index.js', 'dist/debug.js'))
```
menduz: The debug entrypoint must be the same. It only complicate things to test with different files with different side effects

packages/entryPoints/preview.ts
```diff
@@ -1,140 +1,13 @@
...
-import { ETHEREUM_NETWORK, DEBUG, AVOID_WEB3, DEBUG_MESSAGES } from '../config'
-const queryString = require('query-string')
-const qs = queryString.parse(document.location.search)
+import { UnityLoaderType } from '../unity/types'
+import DCLPreview from '../unity/preview/DCLPreview'
 
-type UnityLoaderType = {
-  instantiate(divId: string, manifest: string): UnityGame
-}
+const gameInstance = UnityLoader.instantiate('gameContainer', '/unity/Build/unity.json')
```
menduz, line 17: Classes and side effects are not compatible

menduz (changes requested):
If you are goint to take the "class" approach, it will be necessary to encapsulate side effects in constructors like with

https://github.com/decentraland/explorer/pull/245/files#diff-07f990e64d9429c92f9da52d96cf9ac2R10

Otherwise the code is way harder to follow.

packages/entryPoints/preview.ts
fmiras: Sorry I don't know what do you mean with side effects here

scripts/test.ts
fmiras: The goal of this is to not have exclusive-development code in production, how else do you think we can achieve this? I was also thinking of using config via env variables during compile times so we don't have TLD-based `if`s

packages/unity/preview/DCLPreview.ts
fmiras: I think is better to manage the game instance as an object state rather than a global variable (or actually a whole script-scoped variable) such as `let gameInstance = null` It keeps the entry point code cleaner than before

packages/entryPoints/preview.ts
menduz: unity initialization is a side effect of loading the script because it is not inside any function, constructor or method

if we are going to take the "encapsulation" approach, then the initialization should be part of a constructor or method

scripts/test.ts
menduz: the thing is we don't test production code, we only test one artifact. in what stage we will notice that the positionReporter is not working in the production dist?

packages/unity/preview/DCLPreview.ts
menduz: regarding the side-effect comment, it could be part of UnityPreview class

scripts/test.ts
fmiras: We can still make tests against `index.ts` this `debug.ts` it's mostly for running the explorer within the Unity Editor, I just have to change the tests to keep using index

packages/unity/DCL.ts
fmiras: What's the difference in this case?

packages/entryPoints/preview.ts
eordano: I think that what Agus is trying to say here, is that side-effects need to be tied to the lifecycle of the object of this class, otherwise bad things happen when you try to encapsulate behavior and you don't keep that consistent with the interface you provide.

It's not clear to me that this line injects a `<script>` tag into the `<body>`. Am I right?

fmiras:
Pending fix and test preview entrypoint

PR #248, opened 2019-06-13
Communication auth + fixes

### LINKED ISSUE #56: Deploy new version of the world in dev and fix explorer problems

hugoArregui:
TODO:

- [x] fix comms tests
- [x] re enable preview/not auth (not sure how that works)
- [x] The communication server expects an authentication message soon after the connection is established. The problem is sometimes when you load the client a lot of stuff is going on (download assets and so on) and the auth message is sent too late. Can we move comms initialization later perhaps?.
- [x] We have a comms debug on/off switch, which prints a lot of useful debug information but it's obviously annoying for anyone not interested in comms, I wonder if we can leave this toggle for logging the information but always keep track of it so we can (at least for dev builds), go to the js console and create a dump or enable the log dynamically. Otherwise I think it will be pretty hard to debug the problem of "connection is not working"

Fixes https://github.com/decentraland/world/issues/56


eordano:
Let's leave the toggle on.

eordano (approved):
Works on my machine™

menduz:
Tested against 
- comms (.zone)
- local server (test.ts)
- latest CLI comms server

packages/shared/index.ts
```diff
@@ -64,24 +64,27 @@ async function getAppNetwork(): Promise<ETHEREUM_NETWORK> {
   return net
 }
 
-async function authenticate(): Promise<any> {
-  if (DISABLE_AUTH || PREVIEW) {
```
fmiras, line 68: How do avoid Auth when on development?

packages/shared/index.ts
```diff
@@ -64,24 +64,27 @@ async function getAppNetwork(): Promise<ETHEREUM_NETWORK> {
...
+  let user_id: string
 
-  if (!PREVIEW) {
+  console['group']('connect#login')
+
+  if (PREVIEW) {
+    user_id = 'email|5cdd68572d5f842a16d6cc17'
+  } else {
+    await auth.login(document.getElementsByClassName('loading-image')[0] as HTMLElement)
+    const payload: any = await auth.getAccessTokenData()
```
fmiras, line 78: Is the `any` type necessary here?

fmiras (approved):

packages/shared/index.ts
hugoArregui: we still check for preview, and preview set LOCAL_COMMS whichs disables auth

packages/shared/index.ts
hugoArregui: probably not right now, I think the library has changed and now it returns a well defined type

PR #250, opened 2019-06-13
chore: added auto to transparency mode

AjimenezDCL:
https://github.com/decentraland/unity-client/issues/406

# What? 
Add an AUTO (-1) value to transparencyMode as default

# Why? 
Babylon uses transparencyMode = null to auto set the rendering mode. Adding a -1 value to explicitily allow the enginge to decide what to use.

### DIFF
--- a/packages/decentraland-ecs/src/decentraland/Components.ts
+++ b/packages/decentraland-ecs/src/decentraland/Components.ts
@@ -670,17 +670,18 @@ export class Material extends ObservableComponent {
 
   /**
    * Sets the transparency mode of the material.
-   * Defauts to 0.
+   * Defaults to -1.
    *
    * | Value | Type                                |
    * | ----- | ----------------------------------- |
+   * | -1    | AUTO  (Let the engine decide)       |
    * | 0     | OPAQUE  (default)                   |
    * | 1     | ALPHATEST                           |
    * | 2     | ALPHABLEND                          |
    * | 3     | ALPHATESTANDBLEND                   |
    */
   @ObservableComponent.field
-  transparencyMode?: number
+  transparencyMode?: number = -1
 
   /**
    * Does the albedo texture has alpha?
--- a/packages/decentraland-ecs/src/decentraland/Components.ts
+++ b/packages/decentraland-ecs/src/decentraland/Components.ts
@@ -672,16 +672,16 @@ export class Material extends ObservableComponent {
    * Sets the transparency mode of the material.
    * Defaults to -1.
    *
-   * | Value | Type                                |
-   * | ----- | ----------------------------------- |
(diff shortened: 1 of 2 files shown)


menduz (changes requested):
- It seems trivial, but let the "engine decide" is undefined behavior and we should not foster that.
- What if we set 2 as default value for both scenarios? it will be easier to explain and more robust.
- Every time we use a negative number in enums @eordano and @menduz sacrifice a beautiful kitten

AjimenezDCL:
Using Alpha Blending (2) mode by default it's a terrible idea. Alpha Blending disables Z-Writting and it would break most scenes with Opaque items.

I just made explicit what is being done for babylon right now. If we dont override `transparencyMode` the value stays in Babylon as "null". This ends in Babylon deciding what to do: https://github.com/BabylonJS/Babylon.js/blob/1e92a23f55bcbce7854ed9b290405236779ff6c0/src/Materials/PBR/pbrBaseMaterial.ts#L868

We could do the same in Unity without this PR but then we have inconsiscencies between the default model in `explorer` and the default model in `engine`. It also lead to an inconsistency in the SDK itself since `transparencyMode` will return 0 if not set but internally the transparency mode being used is up to Babylon/Unity

BrianAmadori:
I'd have to agree with @AjimenezDCL here. Actually this is fixing the undefined behaviour, that is leaving the setting to `null`. Maybe we can make the AUTO value 3 or 4 instead -1?.

menduz:
The schema defines `transparencyMode?: number` so if it is missin it is not null, it is the absence of a value (or undefined in JS) what if we assume babylon approach was wrong (because of the delegation of behavior) and remove the optional flag from `transparencyMode` and we decide whatever default value fits best in terms of performance?

Unless the "let the engine decide" has very specific heuristics. In that case I'd search a better name for the default value like "inherit from texture" or whatever the heuristic is

BrianAmadori:
That'd require breaking compatibilty with existing scenes. Because scenes that were previously working would have to set the transparencyMode manually to look correctly.

AjimenezDCL:
Yep, we wanted to avoid having to change current scenes.

Also the heuristic is really specific. Basically is: `hasAlpha ? AlphaBlending : Opaque`

BrianAmadori:
To clarify further: there's not a single value that can fit all cases. Each `transparencyMode` has very specific uses and side effects. You always have to decide if using one or another depending on what you want.

menduz:
> Yep, we wanted to avoid having to change current scenes.

It won't, the current logic is broken

Actionables:
- Keep the logic as it is in this PR, it sounds better than what we have right now
- Change the default -1 to 4, make it the default
- Make it not optional, since it has a defined default value

menduz (approved):

PR #256, opened 2019-06-14
chore: update unity loading screen style

pravusjif:
I updated the loading screen styles to get really close to the following design (minus the tooltip and the loading text percentage)

Screen Shot 2019-06-14 at 16 26 09

### DIFF
--- a/static/tetra.html
+++ b/static/tetra.html
@@ -39,7 +39,7 @@
   </head>
 
   <body class="dcl-loading">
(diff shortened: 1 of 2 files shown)


javiermaria (approved):

menduz:
How is this DCLUnityLoader generated once we update unity's version?

pravusjif:
> How is this DCLUnityLoader generated once we update unity's version?

I just added info on [this issue](https://github.com/decentraland/unity-client/issues/122)'s description to tackle that.

PR #260, opened 2019-06-17
chore: scene loading fixes and performance improvements

eordano:
Drop custom logic from a time when multiple scenes were pushed at the same time.


matiasbargas (dismissed):

eordano:
Despite the fact that I love this change, due to miscommunications with the Unity client unloading scenes when you're too far off, plus lack of communication of offloading events, plus not triggering reloads upon new messages, I found a bug by which there are some scenes that are not reloaded "when you come back". Marked as blocked for the time being

eordano:
Fixed. Made some cleanup of tech debt here and there.

packages/entryPoints/editor.ts
```diff
@@ -78,12 +83,10 @@ async function loadScene(scene: IScene & { baseUrl: string }) {
 }
 
 async function initializePreview(userScene: EnvironmentData<LoadableParcelScene>, parcelCount: number) {
-  loadedParcelSceneWorkers.forEach($ => {
-    $.dispose()
-    loadedParcelSceneWorkers.delete($)
-  })
+  loadedSceneWorkers.forEach($ => forceStopParcelSceneWorker($))
```
menduz, line 86: `forceStopParcelSceneWorker($)` should check `if ($.persistent) return`

packages/shared/world/parcelSceneManager.ts
```diff
@@ -14,41 +15,80 @@ export type EnableParcelSceneLoadingOptions = {
   onLoadParcelScenes?(x: ILand[]): void
 }
 
-export const loadedParcelSceneWorkers: Set<SceneWorker> = new Set()
+export const loadedParcelSceneWorkers: { [base: string]: SceneWorker } = {}
```
menduz: prefer `Map<string, SceneWorker>` over dictionaries

packages/shared/world/parcelSceneManager.ts
```diff
@@ -14,41 +15,80 @@ export type EnableParcelSceneLoadingOptions = {
...
+
+const sceneWorkerByBaseCoordinate: { [base: string]: SceneWorker } = {}
```
menduz: same here

packages/shared/world/parcelSceneManager.ts
```diff
@@ -14,41 +15,80 @@ export type EnableParcelSceneLoadingOptions = {
...
+    worker.parcelScene &&
+    worker.parcelScene.data &&
+    worker.parcelScene.data.data &&
+    worker.parcelScene.data.data.scene &&
+    worker.parcelScene.data.data.scene.base
+  )
+}
+
+export function stopParcelSceneWorker(worker: SceneWorker) {
+  if (worker && !worker.persistent) {
```
menduz, line 54: the name `stopParcelSceneWorker` does not clearly speak about the condition `worker.persistent` since it is a oneLiner I think we can not use those functions, for clarity.

`forceStopParcelSceneWorker` is a good onw

packages/shared/world/parcelSceneManager.ts
```diff
@@ -14,41 +15,80 @@ export type EnableParcelSceneLoadingOptions = {
...
+  const id = getParcelSceneRootCID(parcelScene)
+  if (loadedParcelSceneWorkers[id]) {
+    return loadedParcelSceneWorkers[id]
+  }
+  const parcelSceneWorker = new SceneWorker(parcelScene, transport)
+
+  if (parcelSceneWorker) {
+    loadedParcelSceneWorkers[getParcelSceneRootCID(parcelScene)] = parcelSceneWorker
+
+    if ('sendUserViewMatrix' in parcelSceneWorker) {
```
menduz, line 82: Now that I see it, it should be a worker's responsibility to subscribe/desubscribe to the positionObservable. Can we move that to the constructor/dispose methods to reduce coupling in this module? Up until now I thought it was for loading/unloading life cycle managing

packages/shared/world/parcelSceneManager.ts
```diff
@@ -78,14 +113,9 @@ export async function enableParcelSceneLoading(network: ETHEREUM_NETWORK, option
...
-    const parcelSceneToUnload = await ret.getParcelData(sceneCID)
-    loadedParcelSceneWorkers.forEach($ => {
-      if (!$.persistent && getParcelSceneCID($) === parcelSceneToUnload.mappingsResponse.root_cid) {
-        $.dispose()
-        loadedParcelSceneWorkers.delete($)
-      }
-    })
+  ret.on('Scene.shouldUnload', async (opts: { sceneCID: string }) => {
+    const worker = loadedParcelSceneWorkers[opts.sceneCID]
+    stopParcelSceneWorker(worker)
```
menduz, line 121: what if `!worker`?

packages/unity-interface/dcl.ts
```diff
@@ -282,20 +267,19 @@ async function initializeDecentralandUI() {
 
   await ensureUiApis(worker)
 
-  loadedParcelSceneWorkers.add(worker)
+  loadedParcelSceneWorkers['ui'] = worker
```
menduz: wat

menduz (dismissed):

packages/decentraland-loader/lifecycle/controllers/parcel.ts
```diff
@@ -33,6 +33,7 @@ export class ParcelLifeCycleController extends EventEmitter {
     const sightedParcels = parcelsInScope(this.config, position)
     const sightedParcelsDict: { [key: string]: boolean } = {}
     for (const parcel of sightedParcels) {
+      sightedParcelsDict[parcel] = true
```
fmiras: Wouldn't this array be better as a `Set` type?

packages/decentraland-loader/lifecycle/controllers/scene.ts
```diff
@@ -35,14 +35,16 @@ export class SceneLifeCycleController extends EventEmitter {
   async onSight(position: string) {
     let sceneCID = await this.requestSceneCID(position)
     if (sceneCID) {
-      this.sceneParcelSightCount[sceneCID] = this.sceneParcelSightCount[sceneCID] || 1
+      this.sceneParcelSightCount[sceneCID] = this.sceneParcelSightCount[sceneCID] || 0
+      this.sceneParcelSightCount[sceneCID]++
```
fmiras: I'm seeing that this object is only for counting. Does the state of the class persists when, for example, teleporting? If not we should clear this array right? I suggest using `Map` instead of plain object/array. It's better if we later need to delete elements.

packages/entryPoints/editor.ts
```diff
@@ -121,9 +124,6 @@ async function initializePreview(userScene: EnvironmentData<LoadableParcelScene>
     evtEmitter.emit('entityBackInScene', e)
   })
 
-  // we need closeParcelScenes to enable interactions in preview mode
-  loadedParcelSceneWorkers.add(parcelScene)
```
fmiras, line 125: Is this entrypoint tested with Builder?

packages/shared/world/parcelSceneManager.ts
```diff
@@ -14,41 +15,80 @@ export type EnableParcelSceneLoadingOptions = {
...
+  worker.dispose()
+  const rootCID = getParcelSceneRootCID(worker.parcelScene)
+  const parcelSceneWorker = loadedParcelSceneWorkers[rootCID]
+  delete loadedParcelSceneWorkers[rootCID]
+  const baseCoordinates = getBaseCoordinates(parcelSceneWorker)
+  if (baseCoordinates) {
+    delete sceneWorkerByBaseCoordinate[baseCoordinates]
+  }
+  userViewMatrixWorkers = userViewMatrixWorkers.filter($ => $ !== worker)
+  loadedSceneWorkers = loadedSceneWorkers.filter($ => $ !== worker)
```
fmiras: Same here, if we only want to add and remove elements the `Set` or `Map` structures definitely have better performance + we can use `const` instead of `let` in the definition and work with more accurate data

packages/shared/world/parcelSceneManager.ts
```diff
@@ -114,10 +144,8 @@ export function enablePositionReporting() {
 
   positionObservable.add(obj => {
     worldToGrid(obj.position, position)
-    for (let parcelSceneWorker of loadedParcelSceneWorkers) {
-      if (parcelSceneWorker && 'sendUserViewMatrix' in parcelSceneWorker) {
-        parcelSceneWorker.sendUserViewMatrix(obj)
-      }
+    for (let parcelSceneWorker of userViewMatrixWorkers) {
```
fmiras, line 150: I still don't like using ES6 `for` statements in some places and lambda functions in other places, can we at least write something in dcl standards and explain the "what" and "why". If you tell me those I can write them down.

CC @menduz

fmiras (dismissed):

packages/shared/world/parcelSceneManager.ts
eordano: thanks, will update

packages/shared/world/parcelSceneManager.ts
eordano: Hmm, isn't that a "leaky" abstraction thing? The implementation being a "one liner" shouldn't change the description. I'll add some docs. This is the function that should be used 99% of cases -- "forceStop" is for exceptional things.

eordano: Rationale:
`stopParcelSceneWorker` stops the worker if it's not persistent
`forceStopParcelSceneWorker` doesn't care if it's perssistent

packages/shared/world/parcelSceneManager.ts
eordano: TBH i'd like to keep the purpose of the PR scoped to the critical stuff -- I made changes in how do we do this because I feel it was an improvement in perf but I agree with you...

packages/shared/world/parcelSceneManager.ts
eordano: `stopParcelSceneWorker` handles that case

packages/shared/world/parcelSceneManager.ts
eordano: roger that

packages/decentraland-loader/lifecycle/controllers/scene.ts
eordano: Yes, that is handled by https://github.com/decentraland/explorer/pull/260/files/ca532f4db0b07accaaf0a5f58243d9096216edb7#diff-f986d78a257e5ba8e8887ce2ae54bb6eR44 (you'll get "onSight" and "lostSight" calls. I'll add some tests to clarify.

packages/entryPoints/editor.ts
eordano: I haven't. Will do, thank you for the review

packages/shared/world/parcelSceneManager.ts
eordano: Roger that

packages/shared/world/parcelSceneManager.ts
eordano: Let me know which you prefer. I like these kind of statements, they are more semantic than using lambdas

packages/unity-interface/dcl.ts
eordano: Thanks. should be `loadParcelScene(worker)`

packages/shared/world/parcelSceneManager.ts
menduz: @fmiras `for .. of ..` loops are useful for iterators where the "forEach" does not exist. The performance for critical applications is lower tho. you can also leverage `continue` and `break` statements.

`something.forEach(...)` is usually faster, depending on the free variables of the lamda passed it can be better or worse in performance, counting the frames pushed to the stack.

anyways, analyzing it is a deep dive only worthy for performance critical applications.

TLDR: make it readable and understandable unless it gets called thousands of times

packages/shared/world/parcelSceneManager.ts
```diff
@@ -14,41 +15,80 @@ export type EnableParcelSceneLoadingOptions = {
...
+
+export function loadParcelScene(parcelScene: ParcelSceneAPI, transport?: ScriptingTransport) {
+  const id = getParcelSceneRootCID(parcelScene)
+  if (loadedParcelSceneWorkers[id]) {
+    return loadedParcelSceneWorkers[id]
+  }
+  const parcelSceneWorker = new SceneWorker(parcelScene, transport)
+
+  if (parcelSceneWorker) {
+    loadedParcelSceneWorkers[getParcelSceneRootCID(parcelScene)] = parcelSceneWorker
```
matiasbargas: is this `getParcelSceneRootCID` same as `id` above?

matiasbargas (approved):
1 small comment but I agree with the general idea. Approve waiting for solving all the comments.

eordano:
I changed it to rootcid to explicitly use rootcid. I have no idea in
general if "ID" means location, scene cid, or root cid. I want to start
using scene cid and this is the first step to clean up for that change

packages/entryPoints/editor.ts
eordano: (I see your concern: this is now done by `loadParcelScene` -- it was a leak on the abstraction)

menduz:
Looks good, is the preview for the CLI tested? this code touches code related to the scene reloading

eordano:
> Looks good, is the preview for the CLI tested? this code touches code related to the scene reloading

Reloading is disabled ATM. The current behavior should be the same: if you're asking about the missing `loadedParcelScenes.add` that is now a responsibility of the `loadParcelScene` function.

PS: Yes, I tested the CLI and it works

menduz (approved):

packages/entryPoints/editor.ts
eordano: Confirmed it works as expected

eordano:
@fmiras please review and merge

fmiras (approved):

Issue #257, opened 2019-06-14
Update chat window code to match Romi's new simple design


pravusjif:
<img width="1265" alt="Screen Shot 2019-06-14 at 18 21 07" src="https://user-images.githubusercontent.com/1031741/59538765-40eb9200-8ed1-11e9-8357-fc8ca871b3da.png">

PR #262, opened 2019-06-18
chore: update chat scene

pravusjif:
https://github.com/decentraland/explorer/issues/257

### DIFF
--- a/packages/ui/avatar/chatWindow.ts
+++ b/packages/ui/avatar/chatWindow.ts
@@ -1,14 +1,12 @@
-import { DecentralandInterface, IEvents } from 'decentraland-ecs/src/decentraland/Types'
-import { Color4, Texture, OnEnter } from 'decentraland-ecs/src'
-import { OnTextSubmit, OnBlur, OnChanged, OnClick, OnFocus } from 'decentraland-ecs/src/decentraland/UIEvents'
+import { DecentralandInterface } from 'decentraland-ecs/src/decentraland/Types'
+import { Color4 } from 'decentraland-ecs/src'
+import { OnTextSubmit, OnBlur, OnFocus, OnEnter } from 'decentraland-ecs/src/decentraland/UIEvents'
 
 import {
-  UIImage,
   UIInputText,
   UIText,
   UIContainerStack,
   UIContainerRect,
-  UIFullScreen,
   UIShape,
   UIScrollRect
 } from 'decentraland-ecs/src/decentraland/UIShapes'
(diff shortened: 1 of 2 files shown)


eordano (approved):

packages/ui/avatar/chatWindow.ts
```diff
@@ -120,59 +66,74 @@ const internalState = {
...
+messagesLogScrollContainer.isVertical = false
+messagesLogScrollContainer.isHorizontal = false
+messagesLogScrollContainer.visible = true
+
+const messagesLogStackContainer = new UIContainerStack(messagesLogScrollContainer)
+messagesLogStackContainer.name = 'messages-log-stack-container'
+messagesLogStackContainer.vAlign = 'bottom'
+messagesLogStackContainer.hAlign = 'center'
+// messagesLogStackContainer.adaptHeight = true
+// messagesLogStackContainer.adaptWidth = false
```
fmiras: What's up with this?

packages/ui/avatar/chatWindow.ts
```diff
@@ -182,19 +143,24 @@ export async function initializeChat() {
...
+} */
 
-function setMaximized(maximized: boolean) {
-  container.visible = maximized
-  containerMinimized.visible = !maximized
-  transparentContainer.isVertical = maximized
-  isMaximized = maximized
+function setMaximized(newMaximizedValue: boolean) {
+  if (newMaximizedValue) {
+    if (!isMaximized) {
```
fmiras: `if (newMaximizedValue && !isMaximized)`?

fmiras (changes requested):

packages/ui/avatar/chatWindow.ts
eordano: We're simplifying it -- see the mocks here: https://sketch.cloud/s/2GZ55/a/WweVE2

packages/ui/avatar/chatWindow.ts
pravusjif: If i use `(newMaximizedValue && !isMaximized)`, then i'll have to check the 'complete else' as in `else if (!newMaximizedValue && isMaximized)`. Do you think that's clearer?

packages/ui/avatar/chatWindow.ts
pravusjif: if you referred to the comments I removed them. This PR cleaned up the chat scene code, as there were several objects being created but not used, and the ui elements weren't being parented correctly

packages/ui/avatar/chatWindow.ts
fmiras: Actually you don't because you're asking if `A && !B` and then you're are ensuring that if that whole statement is `false` with the `else if (B)` you already know that A is false.

To clarify: 

```ts
if (newMaximizedValue && !isMaximized) {
  // do stuff
} else if (isMaximized) { // if this is false, then nothing but if is true we already know that first `if` went to false because newMaximizedValue is false, so we don't need the `!newMaximizedValue` 
  // do other stuff
}
// here the stuff if both values are false
```

fmiras (changes requested):

packages/ui/avatar/chatWindow.ts
pravusjif: Great, i just changed it.

fmiras (approved):

PR #263, opened 2019-06-18
fix: correctly report camera position, fix ECS bug

menduz:
Better with https://github.com/decentraland/unity-client/pull/473

### DIFF
--- a/packages/unity-interface/dcl.ts
+++ b/packages/unity-interface/dcl.ts
@@ -37,7 +37,7 @@ import { ensureUiApis } from '../shared/world/uiSceneInitializer'
 import { ParcelIdentity } from '../shared/apis/ParcelIdentity'
 import { IEventNames, IEvents } from '../decentraland-ecs/src/decentraland/Types'
 import { Vector3, Quaternion, ReadOnlyVector3, ReadOnlyQuaternion } from '../decentraland-ecs/src/decentraland/math'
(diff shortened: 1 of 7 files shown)


packages/unity-interface/dcl.ts
```diff
@@ -47,17 +47,19 @@ const preloadedScenes = new Set<string>()
...
+  rotation: Vector3.Zero(),
+  playerHeight: playerConfigurations.height
 }
 
 /////////////////////////////////// HANDLERS ///////////////////////////////////
 
 const browserInterface = {
   /** Triggered when the camera moves */
-  ReportPosition(data: { position: ReadOnlyVector3; rotation: ReadOnlyQuaternion }) {
+  ReportPosition(data: { position: ReadOnlyVector3; rotation: ReadOnlyQuaternion; playerHeight?: number }) {
```
eordano, line 58: Is this likely to change? I'd try to send less data here. This also looks like something that should go in `config` and set only once.

eordano (changes requested):

packages/unity-interface/dcl.ts
menduz: may change with the avatar in runtime

eordano: This is definitely not something that I would send on *every* position update. Please move that to a engine-explorer initialization parameter (or maybe even a configuration value, for the time being).

eordano:
Merging because #473 is merged

PR #264, opened 2019-06-19
feat: lossy queued messages

oOZakMcKrackeNOo:
Tag transform and rendering component updates per entity, and remove older elements from the queue if a new update arrives.

#361


packages/scene-system/scene.system.ts
```diff
@@ -191,13 +191,15 @@ export default class GamekitScene extends Script {
...
             type: 'CreateEntity',
+            id: entityId,
             payload: JSON.stringify({ id: entityId } as CreateEntityPayload)
           })
         },
 
         removeEntity(entityId: string) {
           that.events.push({
             type: 'RemoveEntity',
+            id: entityId,
```
BrianAmadori: If we are sending the `id` like this in all the messages, wouldn't make sense to remove them from the payload to save some json parsing time?

oOZakMcKrackeNOo: Yes, but I left it for later because there will be another round of optimizations (i.e sending only properties that changes, byte arrays, etc.). If you think we should do it anyway, let's do it!

matiasbargas:
related to https://github.com/decentraland/unity-client/pull/477

packages/scene-system/scene.system.ts
BrianAmadori: I'm pretty sure there will be some time before we'll be able to make those changes. I think this optimization would be small enough to fit on the same PR :)

eordano:
Looks great. @menduz any comments?

packages/scene-system/scene.system.ts
```diff
@@ -372,6 +373,7 @@ export default class GamekitScene extends Script {
 
         this.events.push({
           type: 'SceneStarted',
+          id: 'none',
```
menduz: Don't send anything here

packages/scene-system/scene.system.ts
```diff
@@ -301,17 +303,16 @@ export default class GamekitScene extends Script {
         componentDisposed(id: string) {
           that.events.push({
             type: 'ComponentDisposed',
-            payload: JSON.stringify({
-              id
-            } as ComponentDisposedPayload)
+            id: id,
+            payload: JSON.stringify({} as ComponentDisposedPayload)
```
menduz:
```suggestion
            payload: "{}" // JSON of a ComponentDisposedPayload
```

menduz:
The whole point of `{type: string, payload: string}` was to have a high level protocol wrapping/enveloping other packets, this kinda breaks that idea. It makes sense only until we have `id: 'none'`, it is not sound. 

The reason behind that protocol was to have a strategy pattern to parse typed classes depending on the `.type` field of the packet.

If we find a way to solve the `id: 'none'` and the ambiguity of `id` vs `entityId` `componentId` we may proceed. (AttachEntityComponent). If we cannot find a good solution for the model of the high-level wrapper, we can take the information out of the `.payload` right before sending it to unity. But at the cost of more CPU usage

This PR also breaks babylon implementations of client, preview, tests and editor.

oOZakMcKrackeNOo:
> The whole point of `{type: string, payload: string}` was to have a high level protocol wrapping/enveloping other packets, this kinda breaks that idea. It makes sense only until we have `id: 'none'`, it is not sound.
> 
> The reason behind that protocol was to have a strategy pattern to parse typed classes depending on the `.type` field of the packet.
> 
> If we find a way to solve the `id: 'none'` and the ambiguity of `id` vs `entityId` `componentId` we may proceed. (AttachEntityComponent). If we cannot find a good solution for the model of the high-level wrapper, we can take the information out of the `.payload` right before sending it to unity. But at the cost of more CPU usage
> 
> This PR also breaks babylon implementations of client, preview, tests and editor.

I understand why it was implemented this way, but we need to know the component id before parsing the whole payload in order to identify old messages and remove them when a new one has arrived. Let's discuss a better way to achieve this without breaking compatibility with babylon implementation.

menduz:
You can keep the PR as it as it is if you apply the changes to fix babylon and remove the `id: 'none'` by making the id optional

So the envelope may now have the following type

```ts
type EnvelopeMessage = {
  type: string // type of message being carried
  id?: number // id of the object (entity or component) related to the message (optional)
  payload: string // json string of the packet defined by the field .type
}
```

And we need stronger definitions about the package `AttachEntityComponent` about what does `id` mean there. is it entityId or componentId?

eordano:
Won't merge unless the tests pass

eordano:
May I suggest calling this `tag` instead of `id`? In general, I agree with Agus' comments in that there seems to be a mixture of abstraction layers.

Although, (and that's why `tag` becomes more relevant) frequently there are leaks of information between protocols that grow together. I think that the high cost of marshalling/unmarshalling makes this change a must have

packages/scene-system/scene.system.ts
fmiras: YAGNI -> https://martinfowler.com/bliki/Yagni.html

packages/scene-system/scene.system.ts
```diff
@@ -240,10 +242,10 @@ export default class GamekitScene extends Script {
           if (componentNameRE.test(componentName)) {
             that.events.push({
               type: 'AttachEntityComponent',
+              id: id,
```
fmiras: You can just use `id`, is called object property short-hand, info: https://alligator.io/js/object-property-shorthand-es6
```suggestion
              id,
```

packages/scene-system/scene.system.ts
```diff
@@ -289,8 +291,8 @@ export default class GamekitScene extends Script {
           if (componentNameRE.test(componentName)) {
             that.events.push({
               type: 'ComponentCreated',
+              id: id,
```
fmiras:
```suggestion
              id,
```

packages/scene-system/scene.system.ts
```diff
@@ -301,17 +303,16 @@ export default class GamekitScene extends Script {
         componentDisposed(id: string) {
           that.events.push({
             type: 'ComponentDisposed',
-            payload: JSON.stringify({
-              id
-            } as ComponentDisposedPayload)
+            id: id,
```
fmiras:
```suggestion
            id,
```

packages/scene-system/scene.system.ts
```diff
@@ -301,17 +303,16 @@ export default class GamekitScene extends Script {
...
+            payload: JSON.stringify({} as ComponentDisposedPayload)
           })
         },
 
         componentUpdated(id: string, json: string) {
           that.events.push({
             type: 'ComponentUpdated',
+            id: id,
```
fmiras:
```suggestion
            id,
```

fmiras (dismissed):

eordano (approved):

packages/unity-interface/dcl.ts
```diff
@@ -127,12 +127,12 @@ const unityInterface = {
...
   },
-  sendSceneMessage(parcelSceneId: string, method: string, payload: string) {
+  sendSceneMessage(parcelSceneId: string, method: string, payload: string, tag?: string) {
     if (unityInterface.debug) {
       // tslint:disable-next-line:no-console
-      console.log(parcelSceneId, method, payload)
+      console.log(parcelSceneId, method, payload, tag)
     }
-    gameInstance.SendMessage(`SceneController`, `SendSceneMessage`, `${parcelSceneId}\t${method}\t${payload}`)
+    gameInstance.SendMessage(`SceneController`, `SendSceneMessage`, `${parcelSceneId}\t${tag}\t${method}\t${payload}`)
```
menduz: Isn't it better to add things to the tail of the packets when changing protocols? If we do, it may work before updating the unity build

eordano: that's right! thank you

packages/shared/types.ts
```diff
@@ -101,6 +101,7 @@ export type ComponentUpdatedPayload = {
 
 export type EntityAction = {
   type: EntityActionType
+  tag?: string
```
fmiras, line 104: In which cases is this tag `undefined` or missing?

packages/unity-interface/dcl.ts
```diff
@@ -111,12 +111,12 @@ const unityInterface = {
     }
     gameInstance.SendMessage('SceneController', 'LoadParcelScenes', JSON.stringify(parcelsToLoad[0]))
   },
-  sendSceneMessage(parcelSceneId: string, method: string, payload: string) {
+  sendSceneMessage(parcelSceneId: string, method: string, payload: string, tag?: string) {
```
fmiras: In this case I think that `tag` should be optional
```suggestion
  sendSceneMessage(parcelSceneId: string, method: string, payload: string, tag: string) {
```

fmiras (changes requested):

packages/shared/types.ts
menduz: It was when using 'none' but it now uses 'scene' and it makes more sense

packages/unity-interface/dcl.ts
eordano: isn't it?

packages/shared/types.ts
eordano: Tag is optional

eordano: There may be messages without any tag

packages/unity-interface/dcl.ts
menduz: should NOT be optional perhaps?

eordano:
I think optional is fine

menduz:
If you send optional in `sendSceneMessage(a,b,c,tag?: string)` the sent string will end with a `\tundefined`

public/test-parcels/0.0.genetic-experiment/Button.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.fraction = 0
+    this.timeDownLeft = timeDown
+    this.totalTimeDown = timeDown
+  }
+}
+
+export const buttons = engine.getComponentGroup(ButtonData)
+
+export class PushButton implements ISystem {
+  /* update(dt: number) {
```
fmiras, line 24: Please no commented code, If it's for something a TODO task at most with a commit hash reference

public/test-parcels/0.0.genetic-experiment/Environment.ts
```diff
@@ -0,0 +1,39 @@
...
+@Component("environment")
+export class Environment {
+  temperature: number
+  position: Vector3
+  onCreaturesCountUpdated!: any
+  size: number
+
+  creatures: Creature[]
+
+  constructor(temp: number, pos: Vector3, size: number) {
```
fmiras, line 14: tip: by just using `constructor(private temperature: number)` you don't need to do `this.temperature = temp` also, characters are free, no need of use just "temp" it could be confusing

Same with pos and size

packages/unity-interface/dcl.ts
fmiras: Sorry I meant NOT optional

public/test-parcels/0.0.genetic-experiment/Environment.ts
```diff
@@ -0,0 +1,39 @@
...
+    this.temperature = temp
+    this.position = pos
+    this.size = size
+
+    this.creatures = new Array()
+  }
+
+  addCreature(newCreature: Creature) {
+    let newCreaturesCount = this.creatures.push(newCreature)
```
fmiras: use always `const` when defining a varable that is not going to be re-assigned

fmiras:
```suggestion
    const newCreaturesCount = this.creatures.push(newCreature)
```

fmiras (changes requested):

packages/unity-interface/dcl.ts
eordano: No

public/test-parcels/0.0.genetic-experiment/Button.ts
menduz: It is a test scene

public/test-parcels/0.0.genetic-experiment/Environment.ts
menduz: It is a test scene

packages/scene-system/scene.system.ts
```diff
@@ -227,7 +227,7 @@ export default class GamekitScene extends Script {
           if (componentNameRE.test(componentName)) {
             that.events.push({
               type: 'UpdateEntityComponent',
-              tag: entityId,
+              tag: entityId + '_' + classId,
```
menduz, line 230: Ok, this protocol changes don't make sense anymore. 

We should better think about what we want to do and then execute cleanly.

menduz (changes requested):

packages/scene-system/scene.system.ts
oOZakMcKrackeNOo: It's a compromise solution with what we've got at the moment. Right now the Transform entity component is the only lossy message. Next we'll add other components and we have to design a way to manually tell which messages are lossy and which aren't. So maybe the refactor you ask for could be made in that ticket. What do you think?

menduz: I think we should take a more formal process for this issue since the changes went out of our hands. Formalize the requirements in one issue or one pager and discuss it before writing any line of code. AFAIK this problem can also be solved by using protocol buffers and it requires a more radical but better defined change, so this PR will become useless.

oOZakMcKrackeNOo: What we need is to access some of the data while preventing parsing the whole payload just to get them. And that data may vary between messages (for example classId). What do you suggest? Now we're using the tag as a parameters list.

menduz: I suggest to write down the reasons to make this change and then based on those reasons plan and execute with a better idea, this is start from a solution, a hack.

public/test-parcels/0.0.genetic-experiment/Environment.ts
eordano: why do we lower the bar for test scenes?

menduz (approved):

PR #277, opened 2019-06-22
chore: ensure correct artifact building

matiasbargas:


scripts/prepareDist.ts
```diff
@@ -45,19 +48,26 @@ async function copyIndex(filename: string) {
...
   {
     let content = readFileSync(targetIndexHtml).toString()
 
-    if (!content.includes(`${filename}.js`)) {
-      throw new Error(`index.html is dirty and does\'t contain the text "${filename}.js"`)
-    }
+    let didReplace = false
+
+    content = content.replace(new RegExp(fileNameToReplace + '.(S+.)?js'), function(original) {
+      console.log(`>   replacing ${original} -> ${newFileName}`)
```
fmiras:
```suggestion
      console.log(`> replacing ${original} -> ${newFileName}`)
```

scripts/prepareDist.ts
```diff
@@ -45,19 +48,26 @@ async function copyIndex(filename: string) {
     throw new Error(`${targetIndexHtml} does not exist`)
   }
 
-  console.log(`> replace ${filename}.js -> ${newFileName} in html`)
+  console.log(`> replace ${fileNameToReplace}.js -> ${newFileName} in html`)
```
fmiras, line 51: Please be consistent using present continuous verbs or infinitive, but not the two of them

scripts/prepareDist.ts
```diff
@@ -45,19 +48,26 @@ async function copyIndex(filename: string) {
...
   {
     let content = readFileSync(targetIndexHtml).toString()
 
-    if (!content.includes(`${filename}.js`)) {
-      throw new Error(`index.html is dirty and does\'t contain the text "${filename}.js"`)
-    }
+    let didReplace = false
+
+    content = content.replace(new RegExp(fileNameToReplace + '.(S+.)?js'), function(original) {
```
fmiras, line 57: Use arrow function here

fmiras, line 57: I think we should use regex literals to be consistent with the rest of the code

scripts/prepareDist.ts
```diff
@@ -45,19 +48,26 @@ async function copyIndex(filename: string) {
...
 
-  console.log(`> replace ${filename}.js -> ${newFileName} in html`)
+  console.log(`> replace ${fileNameToReplace}.js -> ${newFileName} in html`)
   {
     let content = readFileSync(targetIndexHtml).toString()
 
-    if (!content.includes(`${filename}.js`)) {
-      throw new Error(`index.html is dirty and does\'t contain the text "${filename}.js"`)
-    }
+    let didReplace = false
```
fmiras, line 55: Just compare the already existing variables, you can use `content` for "original" and `result` for the regex match, and if they are the same you already know if it was replaced or not

fmiras (changes requested):

scripts/prepareDist.ts
eordano: I wasn't aware that regexp literals was possible with template `` strings, is it?

fmiras: Just looked into it, you can't so my bad!

PR #280, opened 2019-06-24
fix: setAlias on server

eordano:
# Why?

Otherwise all messages come from peer "0".


menduz (approved):

eordano:
It broke the tests :( seems like something was woking incorrectly before and we weren't noticing

PR #288, opened 2019-06-24
fix: scene messages

menduz:
- Scene messages must reach every scene including the UI, not only parcel scenes in the map
- Chat now works
- Scenes are unloaded
- No scenes are loaded in the same place
- Fixes inconsistencies in several IDs

https://giphy.com/embed/WOYKaXG2xJsBO


eordano (approved):

packages/engine/dcl/index.ts
```diff
@@ -213,12 +213,12 @@ function initDebugCommands() {
...
           if (evt.sourceEvent.key === 'c') {
             colliderVisible = !colliderVisible
-            loadedSceneWorkers.forEach($ =>
+            loadedParcelSceneWorkers.forEach($ =>
               toggleColliderHighlight(colliderVisible, ($.parcelScene as WebGLParcelScene).context.rootEntity)
             )
           } else if (evt.sourceEvent.key === 'b') {
             bboxVisible = !bboxVisible
-            loadedSceneWorkers.forEach($ =>
+            loadedParcelSceneWorkers.forEach($ =>
```
fmiras: I like `loadedSceneWorkers` better :(

packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -7,26 +7,36 @@ import { createLogger } from 'shared/logger'
 const logger = createLogger('loader')
 const { error } = logger
 
-export type SceneMappingResponse = { data: { parcel_id: string; root_cid: string }[] }
+export type DeployedScene = {
+  parcel_id: string
+  root_cid: string
+  /** DO NOT USE THIS ONE YET */
+  scene_cid: ''
```
fmiras, line 14: for what is this for?

packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -7,26 +7,36 @@ import { createLogger } from 'shared/logger'
...
-  }
-  positionToSceneRootCID: Map<string, IFuture<string | null>> = new Map()
+  positionToSceneId: Map<string, IFuture<string | null>> = new Map()
   sceneIdToLandData: Map<string, IFuture<ILand | null>> = new Map()
   rootIdToLandData: Map<string, IFuture<ILand | null>> = new Map()
 
-  constructor(options: { contentServer: string }) {
-    this.options = options
+  constructor(public options: { contentServer: string }) {
+    // stub
```
fmiras, line 31: What's this?

packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -7,26 +7,36 @@ import { createLogger } from 'shared/logger'
...
   }
 
-  async resolveSceneRootCID(pos: string): Promise<string | null> {
-    if (this.positionToSceneRootCID.has(pos)) {
-      return this.positionToSceneRootCID.get(pos)!
+  async resolveSceneSceneId(pos: string): Promise<string | null> {
+    if (this.positionToSceneId.has(pos)) {
+      return this.positionToSceneId.get(pos)!
```
fmiras, line 36: Why the null assertion here? We're already returning a `Promise<string | null>` the user of this function will have to still validate if not null or not

packages/decentraland-loader/lifecycle/controllers/download.ts
menduz: a stub comment. an empty function that needs to exist (constructor) because of our linter doesn't like empty functions

packages/decentraland-loader/lifecycle/controllers/download.ts
menduz: a typing of the server's response. it always returns an empty string

packages/entryPoints/editor.ts
```diff
@@ -185,8 +180,8 @@ export namespace editor {
     return domReadyFuture.isPending ? domReadyFuture : Promise.resolve(canvas)
   }
 
-  export function getScenes() {
-    return loadedSceneWorkers
+  export function getScenes(): Set<SceneWorker> {
+    return new Set(loadedSceneWorkers.values())
```
fmiras, line 184: Cool

fmiras (changes requested):

packages/decentraland-loader/lifecycle/controllers/download.ts
menduz: because it can be resolved to either string | null or throw. null is a valid value

packages/entryPoints/editor.ts
eordano: That defensive programming

PR #291, opened 2019-06-25
refactor: use same logger cross all the project

fmiras:
# What? 
...

# Why? 
...


menduz (changes requested):
please change `log.{method}` by `defaultLogger.{method}`

https://github.com/decentraland/standards/blob/master/standards/react-redux.md#6-bonus-tracks

PR #302, opened 2019-06-28
feat: send user ID via position messages to identify avatars

fmiras:


eordano:
Isn't this the responsibility of the comms server? It could be faked otherwise

menduz:
Checked with Hugo, the alias of every message get's injected by the comms server. It is the id of the connection, cannot be faked. If we wanted the auth0 user ID we need a good user story for it

moliva:
Closing as this will be done on the broker side.

PR #311, opened 2019-07-01
feat: redirect on 401 auth

fmiras:
# What? 
...

# Why? 
...

### DIFF
--- a/packages/shared/index.ts
+++ b/packages/shared/index.ts
@@ -75,20 +75,15 @@ export async function initShared(container: HTMLElement): Promise<ETHEREUM_NETWO
     defaultLogger.log(`Using test user.`)
     user_id = 'email|5cdd68572d5f842a16d6cc17'
   } else {
-    try {
-      await auth.login(container)
-    } catch (error) {
-      defaultLogger.error(error)
-      if (knownTLDs.includes(getTLD())) {
-        window.location.href = `https://explorer.decentraland.${getTLD()}`
-      }
-    }
-
+    await auth.login(container)
     try {
       const payload: any = await auth.getAccessTokenData()
       user_id = payload.user_id
     } catch (e) {
-      console['error'](e)
+      defaultLogger.error(e)
+      if (knownTLDs.includes(getTLD())) {
+        window.location.href = `https://explorer.decentraland.${getTLD()}`
+      }
       console['groupEnd']()
       throw new Error('Authentication error. Please reload the page to try again. (' + e.toString() + ')')
     }
--- a/packages/shared/index.ts
+++ b/packages/shared/index.ts
@@ -3,7 +3,7 @@ import { Auth } from 'decentraland-auth'
 import './apis/index'
 import './events'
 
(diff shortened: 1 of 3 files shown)


packages/shared/index.ts
```diff
@@ -72,32 +72,34 @@ export async function initShared(container: HTMLElement): Promise<ETHEREUM_NETWO
...
       const payload: any = await auth.getAccessTokenData()
       user_id = payload.user_id
     } catch (e) {
-      console['error'](e)
+      defaultLogger.error(e)
+      if (knownTLDs.includes(getTLD())) {
+        window.location.href = `https://explorer.decentraland.${getTLD()}`
+      }
       console['groupEnd']()
       throw new Error('Authentication error. Please reload the page to try again. (' + e.toString() + ')')
```
matiasbargas, line 85: I think we can delete this exception and log it with the previous `defaultLogger.error(e)` instead

fmiras: This error is going to still exists if the application is not hosted at `explorer.decentraland.x` e.g. localhost, new entrypoints or any other thing. If we just log an infinite spinner will take place

eordano: Doesn't this swipe out any `?position=1,1` query parameters?

eordano (changes requested):

packages/shared/index.ts
```diff
@@ -72,32 +72,34 @@ export async function initShared(container: HTMLElement): Promise<ETHEREUM_NETWO
...
     await auth.login(container)
-
     try {
       const payload: any = await auth.getAccessTokenData()
       user_id = payload.user_id
     } catch (e) {
-      console['error'](e)
+      defaultLogger.error(e)
+      if (knownTLDs.includes(getTLD())) {
+        window.location.href = `https://explorer.decentraland.${getTLD()}`
```
menduz: This change should not be handled from here. This is a initShared method, should not handle redirections. Please move the logic to the caller entry point

menduz (changes requested):

packages/unity-interface/initializer.ts
```diff
@@ -45,7 +45,13 @@ const engineInitialized = future()
 export async function initializeUnity(container: HTMLElement): Promise<InitializeUnityResult> {
   _container = container
 
-  await initShared(container)
+  try {
+    await initShared(container)
+  } catch (error) {
+    if (error.message.includes('Authentication error') === '' && knownTLDs.includes(getTLD())) {
+      window.location.href = `https://explorer.decentraland.${getTLD()}${window.location.search}`
```
menduz: Move this one level upper please, to the unity.ts file in the entry points. Here it breaks CLI and Editor

Also `window.location.reload()` makes the trick

menduz (changes requested):

packages/unity-interface/initializer.ts
menduz: https://developer.mozilla.org/en-US/docs/Web/API/Location/reload

menduz (approved):

PR #353, opened 2019-07-23
feat: add 'color' property on NFTShape

pravusjif:


packages/decentraland-ecs/src/decentraland/Components.ts
```diff
@@ -334,6 +334,9 @@ export class NFTShape extends Shape {
   @Shape.readonly
   readonly src!: string
 
+  @ObservableComponent.field
+  color: Color3 = new Color3(0.6404918, 0.611472, 0.8584906) // Light purple as the default background color
```
moliva: Let's create a constant with the default color and set this in the constructor as a default to improve usability.

moliva (changes requested):

moliva (approved):

PR #364, opened 2019-07-25
fix: wait for candidates if the remote peer cannot trickle candidates

menduz:
The GO implementation does not accept interactive initialization, so we have to wait to have all the candidates before sending the answer/offer.

That trait can be detected once we have the remote offer and then query `peer.canTrickleIceCandidates` to wait. This PR adds that logic.


hugoArregui:
Although the initialization is slower for me, it ends up working, I think we can merge this as it is

Screen Shot 2019-07-26 at 09 33 09

I need to make some changes to the protocol to get trickle working, so it may take a while. I think this change + changing the subnet for prod/stg will help a lot

hugoArregui (approved):

PR #371, opened 2019-07-30
update comms protocol, introduce topic identity messages

hugoArregui:


packages/shared/apis/ChatController.ts
```diff
@@ -72,10 +72,11 @@ export class ChatController extends ExposableAPI implements IChatController {
       // If the message was not a command ("/cmdname"), then send message through wire
       const currentUser = getCurrentUser()
       if (!currentUser) throw new Error('cannotGetCurrentUser')
+      if (!currentUser.profile) throw new Error('profileNotInitialized')
       const newEntry = (entry = {
         id: uuid(),
         isCommand: false,
-        sender: currentUser.displayName || currentUser.publicKey || 'unknown',
+        sender: currentUser.profile.name || currentUser.identity || 'unknown',
```
hugoArregui: I suggest to rename identity to user_id, identity makes sense in the context of the broker protocol since it's a library, so the identity of a user may not be a user_id

packages/shared/comms/worldInstanceConnection.ts
```diff
@@ -137,39 +139,35 @@ export class WorldInstanceConnection {
...
-        throw new Error('trying to send a topic message using null unreliable channel')
-      }
+    const message = new TopicIdentityMessage()
+    message.setType(MessageType.TOPIC_IDENTITY)
+    message.setTopic(topic)
+    message.setBody(encodedBody)
 
-      this.connection.sendUnreliable(bytes)
-    }
+    message.setIdentity(encodedUuid)
```
hugoArregui: you don't need to set identity or role, the server will take care of that

hugoArregui (commented):
LGTM

packages/shared/apis/ChatController.ts
```diff
@@ -208,14 +183,14 @@ export class ChatController extends ExposableAPI implements IChatController {
 
       const user = findPeerByName(username)
 
-      if (user && user.publicKey) {
+      if (user && user.identity) {
         // Cannot block yourself
-        if (user.publicKey === currentUser.publicKey) {
+        if (user.identity === currentUser.identity) {
           return { id: uuid(), isCommand: true, sender: 'Decentraland', message: `You cannot block yourself.` }
```
eordano: can we not use this fake `uuid`?

packages/shared/apis/UserIdentity.ts
```diff
@@ -32,8 +32,8 @@ export class UserIdentity extends ExposableAPI implements IUserIdentity {
     if (!user) return null
 
     return {
-      displayName: user.displayName!,
-      publicKey: user.publicKey!
+      displayName: user.profile!.name,
```
eordano: Let's try to be more defensive here!

packages/shared/apis/ChatController.ts
moliva: Good point. Taken.

packages/shared/comms/worldInstanceConnection.ts
moliva: Removed both fields when creating message. Added to test server.

packages/shared/apis/UserIdentity.ts
moliva: Added check here. But from what I understand we need to refactor this interface in the future to enable users to access the identity of the characters and/or certain profile information.

matiasbargas (approved):

PR #378, opened 2019-08-01
Use auth0 instead of decentraland-auth

### LINKED ISSUE #350: migrate login to use Auth0.js
- login should use auth0.js instead of `auth-service`
- use auth-server to get game token. 
- use game token to hit comms, profile and any game server
- use auth0 token to hit claim name server

leohgc:
Close #350 

Ported code from `decentraland-auth` and `auth-service` to interact with auth0


packages/shared/auth/env.ts
```diff
@@ -0,0 +1,14 @@
+const knownTLDs = new Set(['org', 'today', 'zone'])
```
moliva: It might be more suitable to maintain these env functions in `packages/config/index.ts` for the moment.

moliva (approved):

packages/shared/auth/env.ts
leohgc: removed `.env.ts` altogether 😄

PR #410, opened 2019-08-23
fix: global pointer events test

leohgc:
# What? 
...

# Why? 
...


eordano:
Is this still WIP?

packages/decentraland-ecs/src/decentraland/Systems.ts
```diff
@@ -6,6 +6,27 @@ import { ISystem, ComponentAdded, ComponentRemoved, IEntity } from '../ecs/IEnti
 
 declare var dcl: DecentralandInterface | void
 
+/**
+ * @public
+ */
+export class PointerEventSystem implements ISystem {
+  activate(_: Engine) {
```
moliva: `Input` class is already subscribing to these events, we should avoid duplicating this logic.

packages/unity-interface/dcl.ts
```diff
@@ -76,6 +76,13 @@ const browserInterface = {
     }
   },
 
+  pointerUp(event: PointerEvent) {
```
moliva: It might be better to send these events through the `browserInterface#SceneEvent`. In that case we would have the scene id to which this event belongs to as well as avoid having a hit to an entity from another scene.
We would also avoid duplicating the logic to send an event to the scene that is already in place.
What do you think @oOZakMcKrackeNOo ?

PR #421, opened 2019-08-27
feat: use legacy server to fetch avatar data

eordano:


packages/shared/index.ts
```diff
@@ -111,7 +120,17 @@ export async function initShared(container: HTMLElement): Promise<ETHEREUM_NETWO
     }
 
     let spec: ProfileSpec
-    if (response && response.ok) {
+    if (response && !response.ok) {
```
moliva:
```suggestion
    if (!response || !response.ok) {
```

packages/shared/world/profiles.ts
```diff
@@ -15,12 +15,16 @@ export async function resolveProfile(uuid: string = ''): Promise<Profile> {
...
 
   let spec: ProfileSpec
-  // FIXME - forcing random generation until profile migration is finished - moliva - 29/07/2019
-  if (false && response && response.ok) {
+  if (response && response.ok) {
     // @ts-ignore
     spec = (await response.json()) as ProfileSpec
   } else {
-    spec = await createStubProfileSpec()
+    const legacy = await fetchLegacy(uuid)
```
moliva, line 22: We shouldn't do this on `PREVIEW` mode as the user is not logged in.

packages/shared/index.ts
```diff
@@ -111,7 +120,17 @@ export async function initShared(container: HTMLElement): Promise<ETHEREUM_NETWO
...
+      const legacy = await fetch(`${getServerConfigurations().avatar.server}api/profile`, {
```
moliva: Could we use `fetchLegacy` here?

packages/shared/index.ts
eordano: See line 133?

moliva: I mean, what happens if prior to this `fetchProfile` erred and no `response` is set due to an issue. We could still try and fetch from the legacy system.

eordano: Thanks!

PR #430, opened 2019-08-27
feat: support global scene pointer events messaging

oOZakMcKrackeNOo:
Explorer side of pointer down/pointer up global events implementation.

[related to unity-client side](https://github.com/decentraland/unity-client/pull/724)


eordano (approved):

eordano:
Looks good to me!

packages/decentraland-ecs/src/decentraland/Implementation.ts
```diff
@@ -43,11 +43,18 @@ export class DecentralandSynchronizationSystem implements ISystem {
...
 
     this.dcl.onEvent(event => {
+      const data = event.data as any
+
       switch (event.type) {
         case 'uuidEvent':
-          const data = event.data as any
           engine.eventManager.fireEvent(new UUIDEvent(data.uuid, data.payload))
           break
+        case 'pointerDown':
```
BrianAmadori: It makes sense to have `pointerDown` and `pointerUp` as categories in the same level as `uuidEvent`? I feel it'd be better to have just a generic `pointerEvent` here.

eordano: +1

oOZakMcKrackeNOo: But how would you differentiate a pointer down from an up without adding that data to the PointerEvent payload?

Edit: More context - this is a Scene event which type is pointer down/up. So if we put only pointerEvent as its type, we'll need to add that information to the payload or to the event itself.

BrianAmadori: Adding the data to the payload will look better on the long run than having them as separate ids imho.

oOZakMcKrackeNOo: I agree!

eordano (approved):

packages/decentraland-ecs/src/decentraland/Input.ts
```diff
@@ -27,11 +25,11 @@ export type InputState = Record<
 >
 
 /** @public */
-export type LocalPointerEvent = PointerEvent & {
+export type LocalPointerEvent = GlobalInputEventResult & {
   origin: Vector3
   direction: Vector3
   pointer: Pointer
-  hit?: PointerEvent['hit'] & {
+  hit?: GlobalInputEventResult['hit'] & {
```
BrianAmadori, line 32: `hitPoint`, `normal` and `worldNormal` aren't already members of `GlobalInputEventResult['hit']`?

BrianAmadori (approved):

packages/decentraland-ecs/src/decentraland/Input.ts
oOZakMcKrackeNOo: Yes, but I think it's just converting it's types.

PR #431, opened 2019-08-28
chore: refactoring loading/unloading of scenes

moliva:
# What? 
Simplifies the flow of loading/unloading scenes.

# Why? 
The loading/unloading of scenes was thought as independent events based on parcel positions that was prone to errors due to many parcel positions referring to the same scene. This led to having counters and the unloading of scenes was not being performed correctly.

Now, by allowing the `SceneLifeCycleController` to know the whole set of new parcel positions we can fetch scene ids, dedup and actually unload the scenes that are not on sight anymore.


packages/config/index.ts
```diff
@@ -79,7 +79,7 @@ export namespace visualConfigurations {
 }
 
 // Entry points
-export const PREVIEW: boolean = !!(global as any).preview
+export const PREVIEW: boolean = true
```
eordano: oopsie...

eordano (approved):

moliva:
Blocked by https://github.com/decentraland/unity-client/issues/776.

PR #449, opened 2019-08-30
feat: Raycast

eordano:


packages/decentraland-ecs/src/decentraland/Types.ts
```diff
@@ -51,6 +52,10 @@ export type DecentralandInterface = {
   /** set a new parent for the entity */
   setParent(entityId: string, parentId: string): void
 
+  // QUERY
+
+  query(system: string, config: any): void
```
BrianAmadori: The `system`/`config` naming seems iffy to me as we use the word `system` for the ECS systems and its not consistent with the rest of the messages. 
```suggestion
  query(queryId: string, payload: any): void
```

packages/decentraland-ecs/src/decentraland/PhysicsCast.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  public hitAll(ray: Ray, hitCallback: (event: RaycastHitEntities) => void) {
+    const queryId = uuid()
+
+    this.queries[queryId] = hitCallback as (event: RaycastHit) => void
+
+    dcl && dcl.query('raycast', { queryId, queryType: 'HitAll', ray })
+  }
+
+  public hitFirstAvatar(ray: Ray, hitCallback: (event: RaycastHitAvatar) => void) {
```
eordano, line 166: Let's outright remove this before we actually implement it -- could you leave this in another branch? Or does it bring complications when communicating with the renderer?

packages/decentraland-ecs/src/decentraland/Types.ts
eordano: +1

eordano:
Looking good! I can't approve because I started the branch, thanks!

packages/decentraland-ecs/src/decentraland/Events.ts
```diff
@@ -8,6 +8,20 @@ export class UUIDEvent<T = any> {
...
 
+/**
+ * @public
+ */
+@EventConstructor()
+export class RaycastResponse<T> {
+  constructor(
+    public readonly payload: {
+      queryId: string
+      queryType: string
```
moliva, line 19: We could use `QueryType` here instead of accepting any string.

packages/decentraland-ecs/src/decentraland/PhysicsCast.ts
oOZakMcKrackeNOo: We could create another branch with this (feat/avatar-raycast maybe), and then remove all the avatar related stuff.

packages/decentraland-ecs/src/decentraland/Events.ts
oOZakMcKrackeNOo: We can't because that provokes a circular dependence.

moliva (approved):

BrianAmadori (approved):

PR #453, opened 2019-09-02
chore: temporally disable rollbar

eordano:
It's no useful as it is right now:
image


moliva:
What would we need to make it helpful? Do we have created issues for that?

eordano:
Whenever we see an uncatched exception, fix it. Number one offender is:
https://github.com/decentraland/kernel/blob/7df93eb606ed4dd52ee0dd411b0806ae940fe4a7/packages/shared/comms/worldInstanceConnection.ts#L340

PR #460, opened 2019-09-03
fix: show loading screen before teleporting

moliva:
# What? 
Show loading screen before teleporting to avoid showing an empty world prior to the loading screen.


eordano (approved):

eordano:
Shouldn't we leave both? Is it idempotent?

moliva:
In practical terms, the only place where the `onPositionUnsettled` is called nowadays is due to a teleport. Let's leave as is and correct in the future when refactoring.

PR #462, opened 2019-09-03
feat: throttle messages sent

eordano:
Limit to at most 1 each 100ms


marcosnc:
LGTM.
I'll add also a check to avoid sending positions that were too close to previous ones.

moliva (approved):
LGTM.

PR #463, opened 2019-09-04
fix: cache of compiled scenes

eordano:
only cache the particular files that need to be cached


.circleci/config.yml
```diff
@@ -62,23 +62,21 @@ jobs:
...
               make test-scenes
       - save_cache:
           name: Store cached test scenes
           paths:
-            - packages
-            - scripts
-            - static
+            - packages/decentraland-ecs/dist/index.d.ts
+            - static/systems/scene.system.js
             - public
```
moliva: This might incur into inconsistencies with modified scenes. Unfortunately Circle CI doesn't support globs for `save_cache`. How about moving the compiled files to a new location as a next step?

eordano: fixed!

PR #465, opened 2019-09-04
fix: remove scene ready observer when it corresponds

### LINKED ISSUE #437: Scene Lifecycle: 30second timeout is always reached on teleport
Currently, whenever the user teleports, it takes 30 seconds to show the new position. This looks like it's a consequence of Unity not sending the ActivateRenderingACK until the timeout is reached.

moliva:
Closes #437. 

# What? 
Remove scene ready observer only when the event is for that particular scene, otherwise it will never be notified about the readiness of it.


eordano:
Can we add a test for this?

PR #480, opened 2019-09-05
feat: add secure radius before scenes are unloaded

### LINKED ISSUE #341: Differentiate unload/load range for scenes. (M)
Scene loading range should be half (or less) than unloading range to avoid performance hiccups when walking through the world.

moliva:
Fixes #341. 

# What? 
Differentiate load and unload radius to improve experience while exploring the world.

# Why? 
This helps the loading/unloading of scenes, making it less chaotic and easy to handle on the engine side as well.

### DIFF
--- a/packages/config/index.ts
+++ b/packages/config/index.ts
@@ -40,6 +40,7 @@ export namespace parcelLimits {
   export const centimeter = 0.01
 
   export const visibleRadius = 4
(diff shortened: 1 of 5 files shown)


eordano (approved):
Confirmed it works -- didn't look at the code

PR #481, opened 2019-09-06
fix: Auth problems

### LINKED ISSUE #451: Unauthorized User: Login error weirdness on cookies
image (1)

eordano:
Closes https://github.com/decentraland/kernel/issues/451 and https://github.com/decentraland/kernel/issues/473


eordano:
Users are going to see something like the first time if they had previously accessed the world:
image

PR #490, opened 2019-09-09
chore: only throw once on fatal comms error

eordano:
No one is catching this anyways :/


moliva:
Is this happening? Should we be handling this error instead?

eordano:
This happens if the connection is lost for some reason.

It's pretty hard to track all the times when this is called -- pretty much the setPosition and other "hooked-to-global" events.

PR #495, opened 2019-09-11
Fix ice candidates handler: always register it when creating the connection

hugoArregui:
Previously I have set the call of `onCandidate` after the `setRemoteDescription` call thinking the candidates will be buffered in the connection, but after testing it for a while in the server I realise that if the handler is not defined the candidates are simply lost. So, instead, if an ice candidate is gather but no remote description is set, we simply buffer them ourselves and process them all together after we set the remote description


moliva (approved):
LGTM.

PR #496, opened 2019-09-11
feat: notifications in Unity

javiermaria:
Adds the necessary calls that allow kernel to trigger a notification popup inside Unity.
Can be tested using master in unity-client.

### DIFF
--- a/packages/shared/types.ts
+++ b/packages/shared/types.ts
@@ -393,9 +393,9 @@ export type AvatarAsset = {
 }
 
 export enum NotificationType {
-  Generic = 0,
-  ScriptingError,
-  CommsError
+  GENERIC = 0,
+  SCRIPTING_ERROR = 1,
+  COMMS_ERROR = 2
 }
 
 export type Notification = {
--- a/packages/shared/types.ts
+++ b/packages/shared/types.ts
@@ -392,11 +392,17 @@ export type AvatarAsset = {
   }>
 }
 
+export enum NotificationType {
+  Generic = 0,
+  ScriptingError,
+  CommsError
+}
+
 export type Notification = {
-  type: number
+  type: NotificationType
   message: string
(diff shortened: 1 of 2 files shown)


packages/shared/types.ts
```diff
@@ -392,6 +392,15 @@ export type AvatarAsset = {
   }>
 }
 
+export type Notification = {
+  type: number
```
moliva: Could we have an enum according to the corresponding type in Unity?
```   
 public enum NotificationType
    {
        GENERIC,
        SCRIPTING_ERROR,
        COMMS_ERROR,
    }
```

packages/shared/types.ts
```diff
@@ -392,6 +392,15 @@ export type AvatarAsset = {
...
+  message: string
+  buttonMessage: string
+  timer: number
```
moliva: Adding a comment to know the unit of the timer:
```suggestion
  timer: number // in seconds
```

packages/shared/types.ts
```diff
@@ -392,6 +392,15 @@ export type AvatarAsset = {
...
+  scene: string
+  externalCallbackID: string
```
moliva, line 407: Should we always allow for a button action in a notification? Could we maybe add the flexibility to have 0 (i.e. only display information), 1 (e.g. dismiss, ok, default action) or 2 (e.g. ok & cancel, yes & no) options?
The only real use case I can think of now is the comms reconnection retrial, but it might be suitable in the future.

javiermaria: I thought about this, but I believe that there should always be an "OK" or "Dismiss" button so that users can get the message out of their screen. 
The multiple option came up in the Unity side too, and for now will not be implemented, but it would require minor changes.

moliva (approved):

PR #498, opened 2019-09-12
feat: kernel controls which wearables and NFTs are available

javiermaria:
# What? 
...

# Why? 
...


eordano:
Could we have the "NFT" part be inside the description of the wearable? I don't like the idea of adding one extra new entry point for each one new feature we might add to the inventory.

In turn, I suggest we add the "NFT" data inside the profile of the user. We have been talking with Alex about this...

javiermaria:
> Could we have the "NFT" part be inside the description of the wearable? I don't like the idea of adding one extra new entry point for each one new feature we might add to the inventory.
> 
> In turn, I suggest we add the "NFT" data inside the profile of the user. We have been talking with Alex about this...


Sending the available NFTs as part of the profile does make sense (it is the inventory basically). I like having a separate catalog for NFTs. I don't like adding a bool flag to AvatarAssets so that we know if it is an NFT (it's too meta). I see two options here:

**Option 1:**
We specify in which catalog we want to add a Wearable. Since we can only send one parameter, I'd use something like this:
```
export type CatalogCollection = {
  isNFT: boolean
  assets: AvatarAsset[]
}
```

**Option 2:**
- _Normal_ wearables are directly included using the already implemented calls, or similar ones.
- _Available_ NFTs are your inventory, and therefore part of the profile. We store these in Unity in the same type of data structure as the _normal_ wearables. We can change this in the future so that we update the Inventory instead of the full profile.
- _Equipped_ NFTs are part of the AvatarModel in the profile, same as now.

This would update the full profile each time the user acquires/sells one NFT, which could be an issue if their inventory is too large.

eordano (approved):

eordano (approved):

PR #502, opened 2019-09-13
feat: avatar model refactor

eordano:
Combine with #498


packages/decentraland-ecs/src/decentraland/Types.ts
```diff
@@ -344,3 +344,71 @@ export type EngineEvent<T extends IEventNames = IEventNames, V = IEvents[T]> = {
...
+  tags: string[]
+  representations: BodyShapeRespresentation[]
+}
+
+/**
+ * @public
+ */
+export type BodyShapeRespresentation = {
+  bodyShapes: string[]
+  mainFile: string
```
moliva, line 381: Couldn't a wearable's main file differ depending on the body representation (e.g. a black t-shirt for man or woman)?

packages/shared/comms/index.ts
```diff
@@ -223,7 +222,7 @@ export function processProfileMessage(
     const profileVersion = data.getProfileVersion()
 
     peerTrackingInfo.identity = identity
-    peerTrackingInfo.loadProfileIfNecessary(auth, profileVersion)
+    peerTrackingInfo.loadProfileIfNecessary(profileVersion ? parseInt(profileVersion, 10) : 0)
```
moliva, line 228: Might as well use the `undefined` to avoid internal version comparison as we look for any.
```suggestion
    peerTrackingInfo.loadProfileIfNecessary(profileVersion ? parseInt(profileVersion, 10) : undefined)
```

packages/shared/loading/reducer.ts
```diff
@@ -0,0 +1,18 @@
+import { AnyAction } from 'redux'
+import { Events, EventsList, NOT_STARTED } from './types'
+
+export type LoadingState = {
+  status: Events
+}
+export function loadingReducer(state: LoadingState, action: AnyAction) {
```
moliva: Is this as well as `packages/shared/loading/sagas.ts` being used for sth?

packages/shared/passports/sagas.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { expectSaga } from 'redux-saga-test-plan'
```
moliva: Should we move this spec to the `test` folder?

packages/shared/passports/README.md
```diff
@@ -0,0 +1,35 @@
+## Passports
```
moliva, line 1: 👍

packages/ui/avatar/avatarSystem.ts
```diff
@@ -38,21 +38,19 @@ export class AvatarEntity extends Entity {
   }
 
   loadProfile(profile: Profile) {
+    console.log(profile)
```
moliva:
```suggestion
```

packages/ui/avatar/avatarSystem.ts
```diff
@@ -70,6 +68,7 @@ export class AvatarEntity extends Entity {
   }
 
   setUserData(userData: Partial<UserInformation>): void {
+    console.log(userData)
```
moliva:
```suggestion
```

packages/unity-interface/dcl.ts
```diff
@@ -160,8 +158,10 @@ const unityInterface = {
   UnlockCursor() {
     gameInstance.SendMessage('MouseCatcher', 'UnlockCursor')
   },
-  AddWearablesToCatalog(wearables: AvatarAsset[]) {
-    gameInstance.SendMessage('SceneController', 'AddWearablesToCatalog', JSON.stringify(wearables))
+  AddWearablesToCatalog(wearables: Wearable[]) {
+    for (let wearable of wearables) {
```
moliva, line 187: Why send one wearable per call?

packages/decentraland-ecs/src/decentraland/Types.ts
eordano: that's handled by line 377 above (BodyShapeRepresentation is an array).

packages/shared/comms/index.ts
eordano: Kind of, I'm unsure how would that be handled in the protocol buffer?

packages/unity-interface/dcl.ts
eordano: This was a request from the Unity side (cc: @javiermaria ?)

packages/shared/auth/Auth.ts
```diff
@@ -21,35 +18,29 @@ export function isTokenExpired(expiresAt: number) {
 }
 
 export class Auth {
+  ephemeralKey?: BasicEphemeralKey
```
moliva: Avoid making fields public if not needed.

packages/shared/logger.ts
```diff
@@ -17,6 +17,9 @@ export function createLogger(prefix: string): ILogger {
       }
     },
     log(message: string, ...args: any[]): void {
+      if (args && args[0] && args[0].startsWith && args[0].startsWith('The entity is already in the engine.')) {
```
moliva, line 20: We should put in place a better choice for disabling this kind of logs when not in debug mode.

moliva (approved):

packages/shared/logger.ts
eordano: Yeah, the problem with this is that it might be on a deployed parcel already -- so I wouldn't be able to do this in other parts of the code I think

eordano:
deprecated in favor of #553

PR #507, opened 2019-09-17
chore: update scene boundaries test scene

pravusjif:
Replaced tree model usage with an avatar gltf that has several submeshes

### DIFF
--- a/public/test-scenes/0.90.scene-boundaries/game.ts
+++ b/public/test-scenes/0.90.scene-boundaries/game.ts
@@ -142,54 +142,54 @@ sharkVisibilityTrigger.addComponent(
 )
 engine.addEntity(sharkVisibilityTrigger)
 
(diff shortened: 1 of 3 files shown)


public/test-scenes/0.90.scene-boundaries/game.ts
```diff
@@ -142,54 +142,54 @@ sharkVisibilityTrigger.addComponent(
 )
 engine.addEntity(sharkVisibilityTrigger)
 
-// PUSHABLE COLLIDING TREE
-let treeEntity = configureShapeEntityPositions([new Vector3(15, 0, 0)], 0.7, new GLTFShape('models/PalmTree_01.glb'))
```
moliva, line 146: Should we remove the `models/PalmTree_01.glb`?

moliva (approved):

PR #510, opened 2019-09-18
proposed: new pipeline

eordano:
CircleCI is pretty catchy:

https://discuss.circleci.com/t/workflow-job-with-tag-filter-being-run-for-every-commit/20762/4

https://github.com/decentraland/kernel/commit/1b7fb482ecd0b5d5a575dfa1c0f3de6fe754a77e

KISS: better to avoid exclusion rules that are inclusive by default

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -275,60 +275,59 @@ workflows:
   version: 2
   build-deploy:
     jobs:
-      - build:
-          filters:
-            tags:
-              only: /.*/
+      - build
       - hold-explorer-prd:
           type: approval
           requires:
             - build
           filters:
             branches:
-              only: master
-            tags:
-              only: /^\d+\.\d+\.\d(.*)/
+              only: release
       - hold-explorer-stg:
           type: approval
           requires:
             - build
           filters:
             branches:
-              only: master
-            tags:
-              only: /^\d+\.\d+\.\d(.*)/
-      - hold-ecs:
+              only: staging
+      - hold-ecs-prd:
           type: approval
           requires:
             - build
           filters:
             branches:
-              ignore: /.*/
-            tags:
-              only: /^\d+\.\d+\.\d(.*)/
+              only: release
+      - hold-ecs-stg:
+          type: approval
+          requires:
+            - build
+          filters:
+            branches:
+              only: staging
       - deploy-on-pr:
           requires:
             - build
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -283,78 +275,79 @@ workflows:
...
-      - hold-ecs:
+      - hold-ecs-stg:
           type: approval
           requires:
             - build
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
+          filters:
```
moliva:
```suggestion
          filters:
```

.circleci/config.yml
```diff
@@ -283,78 +275,79 @@ workflows:
...
+      - hold-ecs-stg:
           type: approval
           requires:
             - build
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
+          filters:	
+            branches:
```
moliva:
```suggestion
            branches:
```

.circleci/config.yml
```diff
@@ -283,78 +275,79 @@ workflows:
...
           type: approval
           requires:
             - build
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
+          filters:	
+            branches:	
+              only: master
```
moliva:
```suggestion
              only: master
```

.circleci/config.yml
```diff
@@ -283,78 +275,79 @@ workflows:
...
           requires:
             - build
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
+          filters:	
+            branches:	
+              only: master	
+            tags:
```
moliva:
```suggestion
            tags:
```

.circleci/config.yml
```diff
@@ -283,78 +275,79 @@ workflows:
...
             branches:
               only: master
-      - publish-ecs-next:
+      - publish-ecs-prd:
           requires:
             - build
+            - hold-ecs-prd
           filters:
             branches:
               only: master
```
moliva: The issue that I find of running this step only in `master` as opposed to tags is that the `npmPublish` script will not find a tag and therefore will publish the package as a snapshot version (tracked by `next`) as opposed to an actual tagged version in `latest`.
We could think of another way to do this or stick to this way of tagging to publish ECS for the time being.

PR #536, opened 2019-09-25
fix: auth redirectUri

eordano:
Having the parameter was causing a bunch of inconsistencies. This commits causes the auth system to always use the config value


moliva:
With this change we will incur into an issue of dropping query params with which the user tried to log into the world (`ENV`, `position`, etc.). We should at least re add the query parameters to the base location URI.

eordano:
@moliva what do you think about dropping this idea?? please open it again if you think it's still worth it

PR #539, opened 2019-09-25
feat: extract artifacts from a unity build

eordano:


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "6.4.2",
   "license": "Apache-2.0",
   "scripts": {
-    "postinstall": "git lfs install --force",
+    "postinstall": "git lfs install --force && cp node_modules/decentraland-renderer/*.unityweb static/unity/Build/",
```
moliva, line 6:
```suggestion
    "postinstall": "git lfs install --force && cp -r node_modules/decentraland-renderer/Build static/unity/Build/",
```

moliva: Sorry, I understood we wanted everything from the build but now I realize that the JSON config and other files might be something that we want to update manually.

eordano: yes, specially vbecause CachedXMLRequest warnings keep coming up 😅

moliva: I would be concerned by the `unityVersion` mostly. Might only be a label, but we would be losing track of it potentially. 🤔

PR #542, opened 2019-09-25
chore: update profile API

moliva:
# What? 
Updates code to use updated version of profile service

# Why? 
Updates:
- Endpoints
- Response format
- Mappings
- Adds email from Auth0 client

### DIFF
--- a/packages/shared/world/profiles.ts
+++ b/packages/shared/world/profiles.ts
@@ -37,7 +37,7 @@ export async function resolveProfileSpec(uuid: string, spec: ProfileSpec, email?
   return {
     userId: uuid,
     name,
-    email: email || `${name}@decentraland.org`,
+    email: email || '',
     description: spec.description,
     created_at: spec.createdAt,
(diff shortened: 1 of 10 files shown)


packages/shared/world/profiles.ts
```diff
@@ -30,20 +29,19 @@ export async function resolveProfile(accessToken: string, uuid: string = ''): Pr
...
 
-  // TODO - fetch name from claim server - moliva - 22/07/2019
   const name = spec.name || `Guest_${uuid.replace('email|', '').slice(0, 7)}` // strip email| from auth0 uuid
 
   return {
     userId: uuid,
-    name: name,
-    email: `${name}@decentraland.org`, // TODO - populate with actual email when ready - moliva - 29/07/2019
+    name,
+    email: email || `${name}@decentraland.org`,
```
eordano: Since we are at it -- should we drop "@decentraland.org"? This has been confusing people (CC: @xpalaum )

eordano (approved):

PR #553, opened 2019-09-30
feat: avatars new models (big squash)

eordano:
A number of changes in the avatars model and build system.

This includes:

* A new model for avatars
* Starting to test sagas with `redux-saga-test-plan`
* Using `decentraland-renderer`
* Switching all `version` of avatars to be numbers
* Reverting "drop redirectURL from login parameters"
* Pushing catalog to Unity
* `renderer` module to keep track of the renderer's status
* calls to save the user avatar
* TODO: Dropped tests of avatar scene. Consider readding them ASAP


packages/shared/apis/EngineAPI.ts
```diff
@@ -37,7 +35,7 @@ export interface IEngineAPI {
 @registerAPI('EngineAPI')
 export class EngineAPI extends ExposableAPI implements IEngineAPI {
   didStart: boolean = false
-  parcelSceneAPI!: ParcelSceneAPI
+  parcelSceneAPI!: any
```
moliva, line 38: Why change this type?

moliva (approved):

packages/shared/apis/EngineAPI.ts
eordano: There's a circular dependency on this :(

eordano:
Waiting on https://github.com/decentraland/unity-client/pull/760 to merge

PR #554, opened 2019-09-30
fix: copy files on make, not postinstall

eordano:


Makefile
```diff
@@ -53,7 +53,15 @@ docs: $(DECENTRALAND_ECS_TYPEDEF_FILE) ## Generate `decentraland-ecs` documentat
 	@cd $(PWD)/packages/decentraland-ecs; $(PWD)/node_modules/.bin/api-documenter markdown -i types/dcl -o docs
 	@echo "\nAPI Documentation was generated in the packages/decentraland-ecs/docs folder"
 
-build-essentials: $(COMPILED_SUPPORT_JS_FILES) $(DECENTRALAND_ECS_TYPEDEF_FILE) $(BUILD_ECS) $(SCENE_SYSTEM) $(INTERNAL_SCENES) $(DECENTRALAND_LOADER) generate-mocks ## Build the basic required files for the explorer
+RENDERER_BUILD := static/unity/Build/unity.data.unityweb static/unity/Build/unity.wasm.framework.unityweb static/unity/Build/unity.wasm.code.unityweb
+
+static/unity/Build/%.unityweb:
```
moliva: We need this to wire the Unity files in the `static` dir with the `node_modules`.
```suggestion
static/unity/Build/%.unityweb: node_modules/decentraland-renderer/%.unityweb
```

moliva (approved):

Makefile
```diff
@@ -223,8 +231,11 @@ watch: $(SOME_MAPPINGS) build-essentials static/dist/unity.js ## Watch the files
 
 update: ## Update the version of the renderer (Unity build)
```
moliva: How about something like this to force download a new version of the current branch:
```
update:
	@rm -rf node_modules/decentraland-renderer
	@make $(RENDERER_BUILD)
```

Makefile
```diff
@@ -53,7 +53,15 @@ docs: $(DECENTRALAND_ECS_TYPEDEF_FILE) ## Generate `decentraland-ecs` documentat
 	@cd $(PWD)/packages/decentraland-ecs; $(PWD)/node_modules/.bin/api-documenter markdown -i types/dcl -o docs
 	@echo "\nAPI Documentation was generated in the packages/decentraland-ecs/docs folder"
 
-build-essentials: $(COMPILED_SUPPORT_JS_FILES) $(DECENTRALAND_ECS_TYPEDEF_FILE) $(BUILD_ECS) $(SCENE_SYSTEM) $(INTERNAL_SCENES) $(DECENTRALAND_LOADER) generate-mocks ## Build the basic required files for the explorer
+RENDERER_BUILD := static/unity/Build/unity.data.unityweb static/unity/Build/unity.wasm.framework.unityweb static/unity/Build/unity.wasm.code.unityweb
+
+static/unity/Build/%.unityweb:
+	@cp node_modules/decentraland-renderer/`basename $@` static/unity/Build/
```
moliva, line 59: I have to check, but it might probably be better to copy every Unity file, to avoid having inconsistencies in the build.
```suggestion
	@cp node_modules/decentraland-renderer/*.unityweb static/unity/Build/
```

eordano: So, my problem with this was that it makes the `make` command weirdly less effective when running with `-j` (parallel jobs)? In any case, these are "pretty large" files (~13mb total) so I'd rather be cautiously scrappy.

Makefile
eordano: My rationale:

* The right way of doing this is either `npm install decentraland-renderer@THE_VERSION_I_WANT` or modifying the `package.json` accordingly.
* In any of these two cases, either `package.json` is modified (which is a dependency for the file inside `node_modules`), or the file inside `node_modules` is modified, so the `static/unity/...` rule gets triggered by `make watch` (or any other rule that triggers `build-essentials`) anyways. I'd rather kill this `make update` thing.

PR #573, opened 2019-10-10
feat: check for new inventory and send a notification if received

eordano:


moliva (approved):

packages/shared/passports/sagas.ts
```diff
@@ -167,6 +174,15 @@ export function sendWearablesCatalog(catalog: Catalog) {
   window['unityInterface'].AddWearablesToCatalog(catalog)
 }
 
+export function handleNewInventoryItem() {
+  window['unityInterface'].ShowNotification({
+    type: NotificationType.GENERIC,
+    message: 'You received an exclusive wearable NFT mask! Check it out in the avatar editor.',
```
moliva, line 180: It could also happen that the NFT was sold and is no longer available, right?

PR #578, opened 2019-10-15
feat: send performance reports

eordano:


moliva (approved):

packages/unity-interface/initializer.ts
```diff
@@ -81,6 +81,10 @@ namespace DCL {
 
   export function MessageFromEngine(type: string, jsonEncodedMessage: string) {
     if (_instancedJS) {
+      if (type === 'PerformanceReport') {
```
eordano, line 84: @moliva what do you think of this?

moliva: Why should this be a special case? The samples string should come JSON encoded as the rest of the messages from the renderer side or is that an issue?

PR #580, opened 2019-10-15
fix: sending new position to prior position topic when transitioning

### LINKED ISSUE #576: Players seem stuck in their place when in fact they are out of sight
This happens both when the players walk away or teleport from one's position.

This is due to the other player's position not being reported as soon as it moves away and is unsubscribed from the parcel's position topic.

moliva:
Fixes #576 .

### DIFF
--- a/packages/shared/comms/index.ts
+++ b/packages/shared/comms/index.ts
@@ -36,9 +36,9 @@ type Timestamp = number
 type PeerAlias = string
 
 const UNREACHABLE_POSITION: Position = [
-  parcelLimits.parcelSize * 1000,
+  1000 * parcelLimits.parcelSize,
   1000,
-  parcelLimits.parcelSize * 1000,
+  1000 * parcelLimits.parcelSize,
   0,
   0,
   0,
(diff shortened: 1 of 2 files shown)


packages/shared/comms/index.ts
```diff
@@ -35,6 +35,8 @@ import { ProfileForRenderer } from 'decentraland-ecs/src'
 type Timestamp = number
 type PeerAlias = string
 
+const UNREACHABLE_POSITION = [1000, 1000, 1000, 0, 0, 0, 0] as Position
```
eordano: Warning! This falls in parcel `62,62`

eordano (changes requested):
`62,62` is not unreachable

packages/shared/comms/index.ts
moliva: Good catch!

eordano (approved):

packages/shared/comms/index.ts
```diff
@@ -35,6 +35,16 @@ import { ProfileForRenderer } from 'decentraland-ecs/src'
 type Timestamp = number
 type PeerAlias = string
 
+const UNREACHABLE_POSITION: Position = [
+  parcelLimits.parcelSize * 1000,
```
eordano:
```suggestion
  1000 * parcelLimits.parcelSize,
```
Yoda notation might help here?

eordano (approved):
Nit

PR #584, opened 2019-10-17
feat: added call to send the number of new available wearables

javiermaria:
#583


moliva:
This will be used to notify the number of new items as a badge in the avatar hud?

javiermaria:
> This will be used to notify the number of new items as a badge in the avatar hud?

Exactly. I would rather have that information as part of the profile, but that will not be possible until the profile is reviewed and reworked. This will have to do it for now.

moliva (approved):

eordano (approved):

PR #590, opened 2019-10-21
fix: perf report

eordano:


packages/shared/session/getPerformanceInfo.ts
```diff
@@ -6,19 +6,19 @@ export function getPerformanceInfo(samples: string) {
     entries[i] = samples.charCodeAt(i)
     sum += entries[i]
   }
-  entries.sort()
+  const sorted = entries.sort()
```
moliva: `entries.sort()` has side effects, so `entries === sorted`

moliva (approved):

PR #600, opened 2019-10-22
chore: path simulator server

moliva:
# What? 
Simulates a renderer specifying a path through which the character will traverse.

# Why? 
This enables:
- An easier way to test interactions with other avatars/characters/players
- Test communication messages load with kernel side logic
- Further development towards a true E2E client simulator
- A better understanding of kernel <--> renderer message interactions


scripts/runPathSimulator.ts
```diff
@@ -0,0 +1,254 @@
...
+    const next = current + step
+    return next >= target ? target : next
+  },
+  NEGATIVE: step => (current, target) => {
+    const next = current - step
+    return next <= target ? target : next
+  }
+}
+
+function single(...moves) {
```
menduz: nice

scripts/runPathSimulator.ts
```diff
@@ -0,0 +1,254 @@
...
+  return function*() {
+    return { x: 16 * 1000, y: 1000, z: 16 * 1000 }
+  }
+}
+
+function walkTo(x: number, z: number) {
+  return function*() {
+    let position = yield
+
+    const xSense = (x > position.x ? Sense.POSITIVE : Sense.NEGATIVE)(2)
```
eordano: do you think that this might be more legible?
```suggestion
    const moveTowardsTargetX = (x > position.x ? Sense.POSITIVE : Sense.NEGATIVE)(2)
```

eordano (approved):

PR #603, opened 2019-10-23
feat: Implement an alternative message serialization format in Kernel/Unity

oOZakMcKrackeNOo:
# What? 
Implements messaging protocol between kernel and client through ProtoBuffers

# Why? 
Because it's *a lot* faster


packages/shared/stableJson.ts
```diff
@@ -0,0 +1,85 @@
+const json = JSON
+const isArray = Array.isArray
+const objectKeys = Object.keys
+
+export function stableJson({
```
menduz: Why is this necssary?

packages/scene-system/scene.system.ts
```diff
@@ -479,4 +480,38 @@ export default class GamekitScene extends Script {
...
+
+  private generatePBObject(classId: CLASS_ID, json: string): string {
+    let data: string = json
+
+    if (classId === CLASS_ID.TRANSFORM) {
+      const transform: Transform = JSON.parse(json)
+      const pbTransform: PB_Transform = new PB_Transform()
+      const pbPosition: PB_Vector3 = new PB_Vector3()
+      const pbRotation: PB_Quaternion = new PB_Quaternion()
+      const pbScale: PB_Vector3 = new PB_Vector3()
```
menduz: Those can live at top level in the file to avoid instancing new objects every function call

packages/decentraland-ecs/src/decentraland/Implementation.ts
```diff
@@ -174,7 +175,7 @@ export class DecentralandSynchronizationSystem implements ISystem {
       if (isDisposableComponent(component)) {
         this.dcl.attachEntityComponent(event.entity.uuid, event.componentName, getComponentId(component))
       } else if (event.classId !== null) {
-        const componentJson = JSON.stringify(component)
+        const componentJson: string = stableJson(component)
```
menduz: Using stableJson (which is outside of the ECS and may fail because of it) is slower -way- than JSON.stringify

moliva (approved):

packages/unity-interface/dcl.ts
```diff
@@ -348,6 +375,143 @@ class UnityScene<T> implements ParcelSceneAPI {
...
+        message.setScenestarted(new Empty()) // don't know if this is necessary
+        break
+    }
+
+    let arrayBuffer: Uint8Array = message.serializeBinary()
+    return btoa(String.fromCharCode(...arrayBuffer))
+  }
+
+  encodeCreateEntity(createEntityPayload: CreateEntityPayload): PB_CreateEntity {
+    let createEntity: PB_CreateEntity = new PB_CreateEntity()
```
menduz: To reduce allocator & GC overhead, this objects can be created once per file at top level

menduz (changes requested):

PR #620, opened 2019-10-30
fix: inventory notifications

eordano:
fix the dangling notification by storing a value representing "whether or not this item has been notified to the user"


test/unit/passport.saga.test.tsx
```diff
@@ -17,3 +21,30 @@ describe('fetchProfile behavior', () => {
       .run()
   })
 })
+
+describe('notifications behavior', () => {
+  const getReturnsNull = (_: any) => undefined
+  const getReturnsYes = (_: any) => 'sent'
```
moliva: This works because the check is `!getFromDb(storeKey)`, but it is assigned to `notified` instead of `sent` in business code.
```suggestion
  const getReturnsYes = (_: any) => 'notified'
```

packages/shared/passports/sagas.ts
```diff
@@ -339,27 +340,57 @@ export function delay(time: number) {
...
+    const newInventory: string[] = yield select(getInventory, userId)
+    yield call(compareInventoriesAndTriggerNotification, userId, inventory, newInventory)
+  }
+}
+
+export function* compareInventoriesAndTriggerNotification(
+  userId: string,
+  oldInventory: string[],
+  newInventory: string[],
+  getFromDb = getFromLocalStorage,
```
moliva: I would change `getFromDb` to `fetchFromDb` to make it clearer that it's not just a in-memory getter method.

moliva (approved):

packages/shared/passports/sagas.ts
eordano:
```suggestion
  fetchFromDb = getFromLocalStorage,
```

packages/shared/passports/sagas.ts
```diff
@@ -339,27 +340,57 @@ export function delay(time: number) {
...
+  if (areInventoriesDifferent(oldInventory, newInventory)) {
+    const oldItemsDict = oldInventory.reduce(
+      (cumm: Record<WearableId, boolean>, id: string) => ({ ...cumm, [id]: true }),
+      {}
+    )
+    let shouldSendNotification = false
+    for (let item of newInventory) {
+      if (!oldItemsDict[item]) {
+        const storeKey = '__notified_' + item
+        if (!getFromDb(storeKey)) {
```
eordano:
```suggestion
        if (!fetchFromDb(storeKey)) {
```

PR #625, opened 2019-11-02
fix: use take latest for corresponding user id in passport sagas

moliva:
Fixes https://github.com/decentraland/unity-client/issues/1005.

# What? 
Downloads user profiles once per id.


eordano (approved):

eordano:
Awesome work!

PR #627, opened 2019-11-04
fix: limit raycast queries per frame

oOZakMcKrackeNOo:
# What? 
In order to prevent casting many raycasts on the same frame, one possible solution is to make query messages replace older ones each time they arrive to the client. But this approach leads to another problem: sometimes we need to be able to cast more than one ray per frame (i.e. a car that needs to cast three rays on the front to avoid colliding). To allow this, Raycast Instances come to the rescue: a raycast instance is a numeric id that identifies each query that we need to make each frame.

# Why? 
When casting raycasts continuously on each update iteration, several messages can accumulate in client messaging bus. This can lead to performance and user experience issues because raycasts may be calculated many times each frame when not needed. Also, other messages processing are contingent on how many queries are enqueued on the same bus, aggravating the issue.


packages/decentraland-ecs/src/decentraland/PhysicsCast.ts
```diff
@@ -88,12 +88,14 @@ export interface RaycastHitAvatars extends RaycastHit {
...
+  hitAll(ray: Ray, hitCallback: (event: RaycastHitEntities) => void): void
+  hitFirst(ray: Ray, hitCallback: (event: RaycastHitEntity) => void, id: number): void
+  hitAll(ray: Ray, hitCallback: (event: RaycastHitEntities) => void, id: number): void
   /** @internal */
-  hitFirstAvatar: (ray: Ray, hitCallback: (event: RaycastHitAvatar) => void) => void
+  hitFirstAvatar(ray: Ray, hitCallback: (event: RaycastHitAvatar) => void): void
   /** @internal */
-  hitAllAvatars: (ray: Ray, hitCallback: (event: RaycastHitAvatars) => void) => void
+  hitAllAvatars(ray: Ray, hitCallback: (event: RaycastHitAvatars) => void): void
 }
```
moliva: How about this notation instead?
```export interface IPhysicsCast {
  hitFirst(ray: Ray, hitCallback: (event: RaycastHitEntity) => void, id?: number): void
  hitAll(ray: Ray, hitCallback: (event: RaycastHitEntities) => void, id?: number): void
  /** @internal */
  hitFirstAvatar(ray: Ray, hitCallback: (event: RaycastHitAvatar) => void): void
  /** @internal */
  hitAllAvatars(ray: Ray, hitCallback: (event: RaycastHitAvatars) => void): void
}```

File: packages/decentraland-ecs/src/decentraland/PhysicsCast.ts [hunk 2]
```diff
@@ -147,16 +149,16 @@ export class PhysicsCast implements IPhysicsCast {
     return ray
   }
 
-  public hitFirst(ray: Ray, hitCallback: (event: RaycastHitEntity) => void) {
-    const queryId = uuid()
+  public hitFirst(ray: Ray, hitCallback: (event: RaycastHitEntity) => void, id?: number) {
+    const queryId = typeof id === 'number' ? 'rqhf' + id : uuid()
```
moliva: Could we take the `'rqhf'` and `'rqha'` to a QueryType prefix enum?

BrianAmadori (changes requested):
Already requested this change in unity-client. But I'm repeating here: 
in `encodeSceneMessage` under `dcl.ts` we could pre-pack the sceneId+tag in the tag field, avoiding the expensive `StringBuilder` call in unity side.

packages/decentraland-ecs/src/decentraland/PhysicsCast.ts
```diff
@@ -11,6 +11,14 @@ import { Camera } from './Camera'
  */
 export type QueryType = 'HitFirst' | 'HitAll' | 'HitFirstAvatar' | 'HitAllAvatars'
 
+/**
+ * @internal
+ */
+export enum QueryPrefix {
```
moliva: Do we need to export this `enum`? If not needed, let's drop the export.

moliva (approved):

PR #633, opened 2019-11-06
refactor: hud configuration 

javiermaria:
Renderer now handles creation and visibility of HUD elements separately, so kernel must be updated to reflect that

Do not merge until https://github.com/decentraland/unity-client/pull/1138 is merged


moliva (approved):
LGTM.

PR #644, opened 2019-11-10
chore: merge the unity repository into this one

eordano:
Review notes:
- Take special care on reviewing `.circleci`


menduz:
It is publishing the unity package in every PR/build, is that the intended behavior?

eordano:
We already do so -- might change it on a later pr

